PR #3, opened 2022-11-06
Add and Load Chat Stickers

Zintixx:


Maple2.Server.Game/Packets/ChatStickerPacket.cs
```diff
@@ -0,0 +1,45 @@
...
+    public static ByteWriter Load(List<int> favorites, List<ChatSticker> stickers) {
+        var pWriter = Packet.Of(SendOp.ChatStamp);
+        pWriter.Write(Command.Load);
+        pWriter.WriteShort((short) favorites.Count);
+        foreach (int favorite in favorites) {
+            pWriter.WriteInt(favorite);
+        }
+        pWriter.WriteShort((short) stickers.Count);
+        foreach (ChatSticker sticker in stickers) {
+            pWriter.Write(sticker);
```
kOchirasu, line 30: Here and elsewhere, always include the type when using `Write()` (e.g. `Write<ChatSticker>(sticker)`). This avoids potential bugs if the type unexpectedly changes in the future.

The two cases in the file are:
1. `Write<Command>()`
2. `Write<ChatSticker>()`

Maple2.Server.Game/Packets/ChatStickerPacket.cs
```diff
@@ -0,0 +1,45 @@
...
+        Load = 0,
+        ExpiredNotify = 1,
+        Add = 2,
+        Use = 3,
+        GroupChat = 4,
+        Favorite = 5,
+        Unfavorite = 6
+    }
+
+    public static ByteWriter Load(List<int> favorites, List<ChatSticker> stickers) {
```
kOchirasu, line 21: I generally prefer `ICollection<>` here rather than specific implementations (`List<>`). You can still pass lists to `ICollection<>`.

Maple2.Model/Game/ChatSticker.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Runtime.InteropServices;
+
+namespace Maple2.Model.Game;
+
+[StructLayout(LayoutKind.Sequential, Pack = 4, Size = 12)]
+public readonly struct ChatSticker {
+    public int Id { get; init; }
+    public long ExpiryTime { get; init; }
+
+    public ChatSticker(int id, long expiration = long.MaxValue) {
```
kOchirasu, line 10: I generally see 0 used to signify no expiration, is max value actually correct?

Maple2.Server.World/Migrations/20221105211352_ChatSticker.cs
```diff
@@ -0,0 +1,36 @@
...
+            migrationBuilder.RenameColumn(
+                name: "Stamps",
+                table: "character-unlock",
+                newName: "StickerSets");
+
+            migrationBuilder.AddColumn<string>(
+                name: "FavoriteStickers",
+                table: "character-config",
+                type: "json",
+                nullable: false)
```
kOchirasu, line 20: If you're not letting this be nullable, you must specify a `defaultValue: "{}"` otherwise this will become invalid for existing rows.

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -38,4 +45,33 @@ public class ItemUseHandler : PacketHandler<GameSession> {
         
         session.Send(StoryBookPacket.Load(storyBookId));
     }
+
+    private static void HandleChatSticker(GameSession session, Item item)
+    {
+        string? htmlParameter = HttpUtility.HtmlDecode(item.Metadata.Function?.Parameters);
```
kOchirasu, line 51: We should write a util function for parsing these since I assume there are many cases that will need to do the same.

side note, why do we need to decode it as html, I thought it's just xml?

Maple2.Server.Game/Manager/Config/ConfigManager.cs
```diff
@@ -81,6 +84,14 @@ public class ConfigManager {
         }
     }
 
+    public void LoadChatStickers(IDictionary<int, long> stickerSets) {
```
kOchirasu, line 87: You don't need to pass in `stickerSets`, you can retrieve it from `session.Player.Value.Unlock.StickerSets`.

Maple2.Database/Storage/Game/GameStorage.User.cs
```diff
@@ -169,10 +170,10 @@ public partial class Request {
             return Context.TrySaveChanges();
         }
 
-        public (IList<KeyBind>? KeyBinds, IList<QuickSlot[]>? HotBars, List<SkillMacro>?, List<Wardrobe>?, IDictionary<StatAttribute, int>?, SkillBook?) LoadCharacterConfig(long characterId) {
+        public (IList<KeyBind>? KeyBinds, IList<QuickSlot[]>? HotBars, List<SkillMacro>?, List<Wardrobe>?, List<int>? FavoriteStickers, IDictionary<StatAttribute, int>?, SkillBook?) LoadCharacterConfig(long characterId) {
```
kOchirasu, line 173: `FavoriteStickers` is not nullable in db, so it can never be null here? If you want to make it nullable that's fine too but we should be consistent.

Maple2.Server.World/Migrations/20221105211352_ChatSticker.cs
Zintixx: Is there a proper way to set this? Or do I edit this generated migration file?

Maple2.Model/Game/ChatSticker.cs
Zintixx: Yeah, it's max value if there's no expiration.

Maple2.Server.World/Migrations/20221105211352_ChatSticker.cs
kOchirasu: i think you just have to manually set it, that's how ive done it before

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -38,4 +46,46 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
+            if (session.Player.Value.Unlock.StickerSets[stickerSetId] == long.MaxValue) {
+                session.Send(ChatStickerPacket.Error(ChatStickerError.s_msg_chat_emoticon_add_failed_already_exist));
+                return;
+            }
+
+            // does not expire
+            if (duration == 0) {
+                session.Player.Value.Unlock.StickerSets[stickerSetId] = long.MaxValue;
+            } else {
+                session.Player.Value.Unlock.StickerSets[stickerSetId] = Math.Min(session.Player.Value.Unlock.StickerSets[stickerSetId] + duration, long.MaxValue);
```
kOchirasu, line 75: Based on the error message ("- pack has shorter duration than what you have") I don't believe this is the right handling of it. We should be comparing:
1. `session.Player.Value.Unlock.StickerSets[stickerSetId]`
2. `Now + duration`

```cs
long existingTime = 0;
if (session.Player.Value.Unlock.StickerSets.ContainsKey(stickerSetId)) {
    existingTime = session.Player.Value.Unlock.StickerSets[stickerSetId];
    if (existingTime == long.MaxValue) {
        session.Send(ChatStickerPacket.Error(ChatStickerError.s_msg_chat_emoticon_add_failed_already_exist));
        return;
    }
}

long newTime = duration > 0 : DateTimeOffset.UtcNow.ToUnixTimeSeconds() + duration ? long.MaxValue;
if (newTime <= existingTime) {
    session.Send(ChatStickerPacket.Error(ChatStickerError.s_msg_chat_emoticon_add_failed_already_exist));
    return;
}

session.Player.Value.Unlock.StickerSets[stickerSetId] = newTime;
```

Maple2.Database/Model/CharacterConfig.cs
```diff
@@ -18,6 +18,7 @@ internal class CharacterConfig {
     public IList<Wardrobe>? Wardrobes { get; set; }
     public IDictionary<StatAttribute, int>? StatAllocation { get; set; }
     public SkillBook? SkillBook { get; set; }
+    public ISet<int> FavoriteStickers { get; set; }
```
kOchirasu, line 21: the way you make things nullable in the schema is by setting it here:
```cs
public ISet<int>? FavoriteStickers { get; set; }
```

You can regenerate the migration afterwards to make sure it's consistent.

Maple2.Server.Game/Util/XmlParseUtil.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+using System.Xml;
+
+namespace Maple2.Server.Game.Util;
+
+public static class XmlParseUtil {
+    public static Dictionary<string, string> GetParameters(string? xmlString) {
+        Dictionary<string, string> parameters = new Dictionary<string, string>();
+        XmlDocument xmlParameter = new();
```
kOchirasu, line 9: I think you can just do `XmlDocument.LoadXml(...)` which is cleaner.

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -38,4 +46,46 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
+
+        if (!parameters.ContainsKey("id")) {
+            session.Send(ChatStickerPacket.Error(ChatStickerError.s_msg_chat_emoticon_add_failed));
+            return;
+        }
+
+        int stickerSetId = int.Parse(parameters["id"]);
+
+        int duration = 0;
+        if (parameters.TryGetValue(parameters["id"], out string? durationString)) {
```
kOchirasu, line 61: This should be "durationSec" not "id"

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -38,4 +46,46 @@ public class ItemUseHandler : PacketHandler<GameSession> {
         
         session.Send(StoryBookPacket.Load(storyBookId));
     }
+
+    private static void HandleChatSticker(GameSession session, Item item) {
+        Dictionary<string, string> parameters = XmlParseUtil.GetParameters(item.Metadata?.Function?.Parameters);
+
+        if (!parameters.ContainsKey("id")) {
```
kOchirasu, line 53: I would also do the "id" parsing here to avoid crashes:
```cs
if (!parameters.ContainsKey("id") || !int.TryParse(parameters["id"], out int stickerSetId)) {
```

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -38,4 +46,46 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
+            if (duration == 0) {
+                session.Player.Value.Unlock.StickerSets[stickerSetId] = long.MaxValue;
+            } else {
+                session.Player.Value.Unlock.StickerSets[stickerSetId] = Math.Min(session.Player.Value.Unlock.StickerSets[stickerSetId] + duration, long.MaxValue);
+            }
+        } else {
+            if (duration == 0) {
+                session.Player.Value.Unlock.StickerSets[stickerSetId] = long.MaxValue;
+            } else {
+                session.Player.Value.Unlock.StickerSets[stickerSetId] = duration + DateTime.Now.ToEpochSeconds();
```
kOchirasu, line 81: I use `DateTimeOffset.UtcNow.ToUnixTimeSeconds()` everywhere since `DateTime.Now` uses system timezone. There are exceptions when interacting with db, but it's complicated.

Maple2.Database/Storage/Game/GameStorage.User.cs
```diff
@@ -278,6 +280,7 @@ public partial class Request {
                     MaxSkillTabs = 1,
                     ActiveSkillTabId = defaultTab.Id,
                 },
+                FavoriteStickers = new SortedSet<int>(),
```
kOchirasu, line 283: Since this will be nullable, you don't need to set it here.

Maple2.Server.Game/Packets/ChatStickerPacket.cs
```diff
@@ -0,0 +1,54 @@
...
+        ExpiredNotify = 1,
+        Add = 2,
+        Use = 3,
+        GroupChat = 4,
+        Favorite = 5,
+        Unfavorite = 6,
+        Error = 7
+    }
+
+    public static ByteWriter Load(List<int> favorites, ICollection<ChatSticker> stickers) {
```
kOchirasu, line 22: You should also be able to use `ICollection<>` for favorites.

Maple2.Server.Game/Util/XmlParseUtil.cs
Zintixx: If I don't initialize XmlDocument beforehand, it returns `Cannot access non-static method 'LoadXml' in static context`.

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
Zintixx: I feel as though that error is misleading, particularly the 1 of the 3 reasons you pointed out.
`- pack has shorter duration than what you have`
This can't ever really happen. If have an expiring sticker set and you try to use another sticker set that also expires, you end up _adding_ onto the current time.
(i.e. Stickers expire at 12/25. You use a sticker set item that has a duration of 30 days. Your stickers now expire at 1/25)

kOchirasu: I see, maybe this then?
```cs
long existingTime = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
if (session.Player.Value.Unlock.StickerSets.ContainsKey(stickerSetId)) {
    existingTime = session.Player.Value.Unlock.StickerSets[stickerSetId];
    if (existingTime == long.MaxValue) {
        session.Send(ChatStickerPacket.Error(ChatStickerError.s_msg_chat_emoticon_add_failed_already_exist));
        return;
    }
}

long newTime = duration > 0 : existingTime + duration ? long.MaxValue;
if (newTime <= existingTime) {
    session.Send(ChatStickerPacket.Error(ChatStickerError.s_msg_chat_emoticon_add_failed_already_exist));
    return;
}

session.Player.Value.Unlock.StickerSets[stickerSetId] = newTime;
```

PR #4, opened 2022-11-09
Chat Sticker Use, Favorite/Unfavorite & Initial add of Mastery

Zintixx:


Maple2.Model/Metadata/Table/MasteryRecipeTable.cs
```diff
@@ -0,0 +1,25 @@
...
+                        IReadOnlyList<int> RequiredQuests,
+                        long RewardExp,
+                        int RewardMastery,
+                        int HighPropLimitCount,
+                        int NormalPropLimitCount,
+                        IReadOnlyList<Ingredient> RequiredItems,
+                        IReadOnlyList<int> HabitatMapId,
+                        IReadOnlyList<Ingredient> RewardItems);
+
+    public record Ingredient(int ItemId, short Rarity, int Amount, string Tag);
```
kOchirasu, line 21: I will push my change, and you can use the `ItemTag` enum here instead of string. It's weird that they have both ItemId and Tag, perhaps itemId is useless?

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+        if (session.Mastery[entry.Type] < entry.RequiredMastery) {
+            session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_lack_mastery));
+        }
+
+        if (session.Currency.CanAddMeso(-entry.RequiredMeso) != -entry.RequiredMeso) {
+            session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_lack_meso));
+            return;
+        }
+
+        Dictionary<int, IEnumerable<Item>> itemDictionary = new Dictionary<int, IEnumerable<Item>>();
```
kOchirasu, line 58: it's okay to use `var` here since the type is obvious due to the constructor being called.

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+        foreach (MasteryRecipeTable.Ingredient ingredient in entry.RequiredItems) {
+            IEnumerable<Item>? foundItems = session.Item.Inventory.Find(ingredient.ItemId, ingredient.Rarity).ToList();
```
kOchirasu, line 60: I'm not sure this is supposed to be nullable

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+            session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_lack_mastery));
+        }
+
+        if (session.Currency.CanAddMeso(-entry.RequiredMeso) != -entry.RequiredMeso) {
+            session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_lack_meso));
+            return;
+        }
+
+        Dictionary<int, IEnumerable<Item>> itemDictionary = new Dictionary<int, IEnumerable<Item>>();
+        foreach (MasteryRecipeTable.Ingredient ingredient in entry.RequiredItems) {
```
kOchirasu, line 59: At some point we need to create a common utility for this. I have done it already for gem upgrading, and have repeated it again for enchanting. Something like [this](https://github.com/kOchirasu/Maple2/blob/master/Maple2.Server.Game/PacketHandlers/ItemSocketHandler.cs#L389-L429). I have a better version of this locally, i'll try to push it soon.

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
```diff
@@ -0,0 +1,111 @@
...
+            }
+        }
+
+        foreach (MasteryRecipeTable.Ingredient rewardItem in entry.RewardItems) {
+            if (!ItemMetadata.TryGet(rewardItem.ItemId, out ItemMetadata? itemMetadata)) {
+                continue;
+            }
+            Item item = new(itemMetadata, rewardItem.Rarity, rewardItem.Amount);
+            if (!session.Item.Inventory.CanAdd(item)) {
+                // TODO: Mail to player
```
kOchirasu, line 104: What I want instead of some kind of "SafeAdd" on inventory, so that items that cannot be added will automatically be sent to mail. I think there are lots of cases where this is needed.

Maple2.File.Ingest/Mapper/TableMapper.cs
```diff
@@ -6,6 +6,7 @@
 using Maple2.File.Parser.Xml.Table;
 using Maple2.Model.Enum;
 using Maple2.Model.Metadata;
+using Z.BulkOperations;
```
kOchirasu, line 9: I think this was added by mistake.

Maple2.Model/Game/User/Mastery.cs
```diff
@@ -1,10 +1,14 @@
...
+using Maple2.Tools;
+using Maple2.Tools.Extensions;
 
 namespace Maple2.Model.Game; 
 
-[StructLayout(LayoutKind.Sequential, Pack = 4, Size = 48)]
-public struct Mastery {
+//[StructLayout(LayoutKind.Sequential, Pack = 4, Size = 48)]
+public class Mastery : IByteSerializable {
     private int Unknown;
```
kOchirasu, line 10: This was mostly needed as a placeholder for struct serialization. Since you are use `IByteSerializable` now, you can just have a `WriteInt();` for this.

Maple2.Model/Error/MasteryError.cs
```diff
@@ -0,0 +1,18 @@
...
+    s_mastery_error_lack_mastery = 1,
+    [Description("Not enough mesos.")]
+    s_mastery_error_lack_meso = 2,
+    [Description("You have not completed the required quest.")]
+    s_mastery_error_lack_quest = 3,
+    [Description("Not enough items.")]
+    s_mastery_error_lack_item = 4,
+    [Description("Insufficient level.")]
+    s_mastery_error_invalid_level = 7,
+}
```
kOchirasu, line 18: Under Settings > General in Rider you can add these settings to add newlines at ends of files:
image

kOchirasu (commented):
For the future, it would be better if you can split distinct changes into different pull requests. It makes things easier to review. (e.g. Sticker PR separate from Mastery PR)

Maple2.Model/Metadata/Table/MasteryRecipeTable.cs
Zintixx: itemId is useless, yeah. if a tag is present, it uses that. I suppose itemId it's really just for display purposes.

kOchirasu: I think we can skip parsing it then. I wonder if `rarity` is the unknown field in [IngredientInfo](https://github.com/kOchirasu/Maple2/blob/master/Maple2.Model/Game/Item/IngredientInfo.cs)

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
kOchirasu: No action here, just FYI.

Maple2.Model/Metadata/Table/MasteryRecipeTable.cs
Zintixx: In case of mastery recipes, we should still parse it. There are some ingredients that _dont_ use ItemTags.
What uses IngredientInfo?

Maple2.Server.Game/PacketHandlers/ChatStickerHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        session.Send(ChatStickerPacket.Unfavorite(stickerId));
+    }
+
+    private bool TryUseSticker(Unlock unlock, int stickerId) {
+        if (!TableMetadata.ChatStickerTable.Entries.TryGetValue(stickerId, out ChatStickerMetadata? metadata)) {
+            return false;
+        }
+
+        return unlock.StickerSets.ContainsKey(metadata.GroupId) && unlock.StickerSets[metadata.GroupId] >= DateTime.Now.ToEpochSeconds();
```
kOchirasu, line 82: `DateTimeOffset.UtcNow.ToUnixTimeSeconds()`

Maple2.Model/Error/MasteryError.cs
```diff
@@ -0,0 +1,18 @@
+﻿// ReSharper disable InconsistentNaming, IdentifierTypo
+
+using System.ComponentModel;
+
+namespace Maple2.Model.Error;
+
+public enum MasteryError : short {
```
kOchirasu, line 7: s_mastery_error_unknown = 5
s_anti_addiction_cannot_receive = 12

Maple2.Server.Game/Packets/MasteryPacket.cs
```diff
@@ -0,0 +1,58 @@
...
+        pWriter.WriteShort((short) item.Rarity);
+
+        return pWriter;
+    }
+    
+    public static ByteWriter GetCraftedItem(MasteryType type, MasteryRecipeTable.Ingredient item) {
+        var pWriter = Packet.Of(SendOp.Mastery);
+        pWriter.Write<Command>(Command.GetCraftedItem);
+        pWriter.WriteShort((short) type);
+        pWriter.WriteInt(item.Amount);
```
kOchirasu, line 44: should be `ICollection<Item>` instead and loop over it
image

Maple2.Server.Game/Packets/MasteryPacket.cs
```diff
@@ -0,0 +1,58 @@
...
+        pWriter.WriteInt(); // unk
+
+        return pWriter;
+    }
+    
+    public static ByteWriter ClaimReward(int rewardBoxDetails, Item item) {
+        var pWriter = Packet.Of(SendOp.Mastery);
+        pWriter.Write<Command>(Command.ClaimReward);
+        pWriter.WriteInt(rewardBoxDetails);
+        pWriter.WriteInt(item.Amount);
```
kOchirasu, line 33: here as well:
image

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+                session.Item.Inventory.Find(ingredient.ItemId, ingredient.Rarity).ToList();
+            int totalCount = foundItems.Sum(item => item.Amount);
+            if (totalCount < ingredient.Amount) {
+                session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_lack_item));
+                return;
+            }
+            itemDictionary.Add(ingredient.ItemId, foundItems);
+        }
+
+        if (entry.RequiredItems.Any(ingredient => session.Item.Inventory.Find(ingredient.ItemId, ingredient.Rarity).ToList().Count < ingredient.Amount)) {
```
kOchirasu, line 70: I think this is doing the wrong thing, it is counting the number of stacks of the item rather than actual item amount. See above rewrite.

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+        var itemDictionary = new Dictionary<int, IEnumerable<Item>>();
+        foreach (MasteryRecipeTable.Ingredient ingredient in entry.RequiredItems) {
+            IEnumerable<Item> foundItems = ingredient.Tag != ItemTag.None ? session.Item.Inventory.Filter(item => item.Metadata.Property.Tag == ingredient.Tag) :
+                session.Item.Inventory.Find(ingredient.ItemId, ingredient.Rarity).ToList();
+            int totalCount = foundItems.Sum(item => item.Amount);
+            if (totalCount < ingredient.Amount) {
+                session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_lack_item));
+                return;
+            }
+            itemDictionary.Add(ingredient.ItemId, foundItems);
```
kOchirasu, line 67: I think this is doing the wrong thing, you are always indexing by itemId, but some ingredients use tag instead. See above rewrite.

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+        if (session.Mastery[entry.Type] < entry.RequiredMastery) {
+            session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_lack_mastery));
+        }
+
+        if (session.Currency.CanAddMeso(-entry.RequiredMeso) != -entry.RequiredMeso) {
+            session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_lack_meso));
+            return;
+        }
+
+        var itemDictionary = new Dictionary<int, IEnumerable<Item>>();
```
kOchirasu, line 58: Rewrite using the same pattern I used for gemstones. (I did not test this so I don't even know if it compiles, but you get the idea)

```cs
// Build this index so we don't need to find materials twice.
Dictionary<int, IList<Item>> materialsById = entry.RequiredItems.ToDictionary(
    entry => entry.ItemId,
    entry => session.Item.Inventory.Find(entry.ItemId, entry.Rarity).ToList()
);
Dictionary<ItemTag, IList<Item>> materialsByTag = entry.RequiredItems.ToDictionary(
    entry => entry.Tag,
    entry => session.Item.Inventory.Filter(item => item.Metadata.Property.Tag == entry.Tag)
);

lock (session.Item) {
    foreach (MasteryRecipeTable.Ingredient ingredient in entry.RequiredItems) {
        int remaining = ingredient.Amount;
        if (ingredient.Tag != ItemTag.None) {
            foreach (Item material in materialsByTag[ingredient.Tag]) {
                remaining -= material.Amount;
                if (remaining <= 0) {
                    break;
                }
            }
        } else {
            foreach (Item material in materialsById[ingredient.ItemId]) {
                remaining -= material.Amount;
                if (remaining <= 0) {
                    break;
                }
            }
        }

        if (remaining > 0) {
            session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_lack_item));
            return;
        }
    }

    foreach (MasteryRecipeTable.Ingredient ingredient in entry.RequiredItems) {
        int remaining = ingredient.Amount;
        if (ingredient.Tag != ItemTag.None) {
            foreach (Item material in materialsByTag[ingredient.Tag]) {
                int consume = Math.Min(remaining, material.Amount);
                if (!session.Item.Inventory.Consume(material.Uid, consume)) {
                    Logger.Fatal("Failed to consume item {ItemUid}", material.Uid);
                    throw new InvalidOperationException($"Fatal: Consuming item: {material.Uid}");
                }
    
                remaining -= consume;
                if (remaining <= 0) {
                    break;
                }
            }
        } else {
            foreach (Item material in materialsById[ingredient.ItemId]) {
                int consume = Math.Min(remaining, material.Amount);
                if (!session.Item.Inventory.Consume(material.Uid, consume)) {
                    Logger.Fatal("Failed to consume item {ItemUid}", material.Uid);
                    throw new InvalidOperationException($"Fatal: Consuming item: {material.Uid}");
                }
    
                remaining -= consume;
                if (remaining <= 0) {
                    break;
                }
            }
        }
    }
}
```

Zintixx: Wouldn't this count ingredients twice if there is a matching item id and tag?
I think I get it tho. let me see what I can do

kOchirasu: It would ignore item id and only use tag.

PR #6, opened 2022-11-11
Title use scrolls and changing player titles

Zintixx:


Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -86,4 +90,20 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
         session.Send(ChatStickerPacket.Add(item, new ChatSticker(stickerSetId, session.Player.Value.Unlock.StickerSets[stickerSetId])));
     }
+
+    private static void HandleTitleScroll(GameSession session, Item item) {
+        if (!int.TryParse(item.Metadata?.Function?.Parameters, out int titleId)) {
+            return;
+        }
+
+        if (session.Player.Value.Unlock.Titles.Contains(titleId)) {
+            session.Send(NoticePacket.Message("s_title_scroll_duplicate_err"));
```
kOchirasu, line 100: You should be able to use the `StringCode` directly. There's an implicit conversion to handle it properly.

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -86,4 +90,20 @@ public class ItemUseHandler : PacketHandler<GameSession> {
 
         session.Send(ChatStickerPacket.Add(item, new ChatSticker(stickerSetId, session.Player.Value.Unlock.StickerSets[stickerSetId])));
     }
+
+    private static void HandleTitleScroll(GameSession session, Item item) {
+        if (!int.TryParse(item.Metadata?.Function?.Parameters, out int titleId)) {
```
kOchirasu, line 95: `item.Metadata` is not nullable

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -86,4 +90,20 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
+            return;
+        }
+
+        if (session.Player.Value.Unlock.Titles.Contains(titleId)) {
+            session.Send(NoticePacket.Message("s_title_scroll_duplicate_err"));
+            return;
+        }
+
+        session.Player.Value.Unlock.Titles.Add(titleId);
+        session.Item.Inventory.Consume(item.Uid, 1);
```
kOchirasu, line 105: don't we need to validate that Consume succeeded before making changes?

Maple2.Server.Game/PacketHandlers/UserEnvHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+            case Command.ChangeTitle:
+                HandleChangeTitle(session, packet);
+                break;
+        }
+    }
+
+    private void HandleChangeTitle(GameSession session, IByteReader packet) {
+        int titleId = packet.ReadInt();
+
+        if (titleId < 0 || !session.Player.Value.Unlock.Titles.Contains(titleId) && titleId != 0) {
```
kOchirasu, line 28: I think this can be simplified to:
```cs
titleId <= 0 || !session.Player.Value.Unlock.Titles.Contains(titleId)
```

but maybe just doing the Contains check is sufficient. Why would we be adding invalid titleIds to a character.

Maple2.Server.Game/Packets/UserEnvPacket.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System.Collections.Generic;
+using Maple2.Model.Error;
+using Maple2.Model.Game;
+using Maple2.PacketLib.Tools;
+using Maple2.Server.Core.Constants;
+using Maple2.Server.Core.Packets;
+
+namespace Maple2.Server.Game.Packets;
+
+public static class UserEnvPacket {
```
kOchirasu, line 10: In this file, fix braces

Maple2.Server.Game/PacketHandlers/UserEnvHandler.cs
Zintixx: A titleId of 0 is valid (No title). But I'll adjust so it only checks for 0 or contains.

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -86,4 +90,20 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
+        if (!int.TryParse(item.Metadata.Function?.Parameters, out int titleId)) {
+            return;
+        }
+
+        if (session.Player.Value.Unlock.Titles.Contains(titleId)) {
+            session.Send(ChatPacket.Alert(StringCode.s_title_scroll_duplicate_err));
+            return;
+        }
+
+        session.Player.Value.Unlock.Titles.Add(titleId);
```
kOchirasu, line 104: This should also happen after consuming, otherwise you could get a title without using the scroll or whatever.

Maple2.Server.Game/Packets/UserEnvPacket.cs
```diff
@@ -0,0 +1,44 @@
...
+        pWriter.WriteInt(titleId);
+        return pWriter;
+    }
+
+    public static ByteWriter LoadTitles(ISet<int> titles) {
+        var pWriter = Packet.Of(SendOp.UserEnv);
+        pWriter.Write<Command>(Command.LoadTitles);
+        pWriter.WriteInt(titles.Count);
+        foreach (int title in titles)
+        {
```
kOchirasu, line 38: nit: braces

Maple2.Server.Game/PacketHandlers/UserEnvHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+            case Command.ChangeTitle:
+                HandleChangeTitle(session, packet);
+                break;
+        }
+    }
+
+    private void HandleChangeTitle(GameSession session, IByteReader packet) {
+        int titleId = packet.ReadInt();
+
+        if (!session.Player.Value.Unlock.Titles.Contains(titleId) && titleId != 0) {
```
kOchirasu, line 28: nit: check `titleId != 0` first to short-circuit this check since that's the cheaper & more common case.

PR #7, opened 2022-11-12
Check Character Name on Creation

Zintixx:


Maple2.Server.Login/PacketHandlers/CharacterManagementHandler.cs
```diff
@@ -104,6 +104,22 @@ private enum Command : byte {
         var jobCode = packet.Read<JobCode>();
         var job = (Job) ((int)jobCode * 10);
         string name = packet.ReadUnicodeString();
+        
+        if (name.Length < 2) {
```
kOchirasu, line 108: Add a constant: `CharacterNameLengthMin = 2`

Maple2.Server.Login/PacketHandlers/CharacterManagementHandler.cs
```diff
@@ -104,6 +104,22 @@ private enum Command : byte {
...
+            session.Send(CharacterListPacket.CreateError(s_char_err_name));
+            return;
+        }
+        
+        if (name.Length > 12) {
```
kOchirasu, line 113: `Constant.CharacterNameLengthMax`

I would also just choose one of the errors to return here rather than do nothing.

Maple2.Server.Login/PacketHandlers/CharacterManagementHandler.cs
```diff
@@ -14,6 +14,7 @@
 using Maple2.Server.Core.Packets;
 using Maple2.Server.Login.Session;
 using Maple2.Server.World.Service;
+using Microsoft.VisualBasic;
```
kOchirasu, line 17: remove

PR #8, opened 2022-11-13
Buddy Badge

Zintixx:


Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -36,12 +38,21 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
+            case ItemFunction.OpenCoupleEffectBox:
+                HandleBuddyBadgeBox(session, packet, item);
+                break;
             default:
                 Logger.Warning("Unhandled item function: {Name}", item.Metadata.Function?.Type);
                 return;
         }
     }
 
+    #region Autofac Autowired
```
kOchirasu, line 50: nit: this region belongs above the `Handle()` function

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -106,4 +117,91 @@ public class ItemUseHandler : PacketHandler<GameSession> {
             session.Player.Value.Unlock.Titles.Add(titleId);
         }
     }
+
+    private void HandleBuddyBadgeBox(GameSession session, IByteReader packet, Item item) {
+        string targetUser = packet.ReadUnicodeString();
+
+        if (targetUser == session.Player.Value.Character.Name) {
+            session.Send(NoticePacket.MessageBox(new InterfaceText(StringCode.s_couple_effect_error_openbox_myself_char)));
```
kOchirasu, line 125: Here and below, there is an implicit constructor for `InterfaceText` from `StringCode` so you can just pass `StringCode` in.

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -106,4 +117,91 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
+            },
+        };
+
+        receiverMail.SetTitle(StringCode.s_couple_effect_mail_title_receiver);
+        receiverMail.SetContent(StringCode.s_couple_effect_mail_content_receiver);
+        receiverMail.SetSenderName(StringCode.s_couple_effect_mail_sender);
+
+        receiverMail = db.CreateMail(receiverMail);
+        if (receiverMail == null) {
+            session.Send(NoticePacket.MessageBox(new InterfaceText(StringCode.s_couple_effect_error_openbox_unknown)));
```
kOchirasu, line 182: Let's throw an exception here for now because this should never happen as the item was already consumed.

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -106,4 +117,91 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
+            return;
+        }
+
+        using GameStorage.Request itemDb = session.GameStorage.Context();
```
kOchirasu, line 186: You don't need a separate db connection, just use the same one from above. (Actually just realized this is unused, so remove it)

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -106,4 +117,91 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
+        Item? receiverItem = db.CreateItem(receiverMail.Id,
+            new(itemMetadata, buddyBadgeBoxParams[1]) {
```
kOchirasu, line 188: nit: `new Item()` instead of `new()`

Maple2.Server.Game/PacketHandlers/BuddyBadgeHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    private void HandleStop(GameSession session, IByteReader packet) {
+        long characterId = packet.ReadLong();
+
+        session.Field?.Broadcast(BuddyBadgePacket.Stop(characterId), session);
+    }
+
+    private bool HasBuddyBadgeEquipped(GameSession sender, GameSession receiver) {
```
kOchirasu, line 56: I think you can simplify this quite a bit:
```cs
if (!sender.Item.Equips.Badge.TryGetValue(BadgeType.Buddy, out Item? senderBadge)) {
    return false;
}
if (!receiver.Item.Equips.Badge.TryGetValue(BadgeType.Buddy, out Item? receiverBadge)) {
    return false;
}

return senderBadge.Id == receiverBadge.Id && ...
```

This way you don't need to constantly perform lookups in the dictionary. While cheap, it is not free to do so.

Maple2.Server.Game/Manager/Items/EquipManager.cs
```diff
@@ -174,19 +174,19 @@ public class EquipManager {
...
+            item.Group = ItemGroup.Badge;
             item.Slot = -1;
             session.Field?.Broadcast(EquipPacket.EquipBadge(session.Player, item));
 
             if (item.Badge.Type == BadgeType.PetSkin) {
                 session.Pet?.BadgeChanged(item.Badge);
             }
 
-            return true;
+            return Badge.TryAdd(item.Badge.Type, item);
```
kOchirasu, line 189: Rather than making this call which could fail, could you follow the pattern used for gear equipping:
```cs
Badge[item.Badge.Type] = item;
```

Maple2.Server.Game/PacketHandlers/BuddyBadgeHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+        if (!sender.Item.Equips.Badge.TryGetValue(BadgeType.Buddy, out Item? senderBadge)) {
+            return false;
+        }
+        if (!receiver.Item.Equips.Badge.TryGetValue(BadgeType.Buddy, out Item? receiverBadge)) {
+            return false;
+        }
+
+        return receiverBadge.Id != senderBadge.Id ||
+               receiverBadge.CoupleInfo?.Name != receiver.Player.Value.Character.Name ||
+               receiverBadge.CoupleInfo?.CharacterId != receiver.Player.Value.Character.Id ||
```
kOchirasu, line 64: shorthand for `receiver.Player.Value.Character.Id` is `receiver.CharacterId`, same for sender

Maple2.Server.Game/PacketHandlers/BuddyBadgeHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    private static bool HasBuddyBadgeEquipped(GameSession sender, GameSession receiver) {
+        if (!sender.Item.Equips.Badge.TryGetValue(BadgeType.Buddy, out Item? senderBadge)) {
+            return false;
+        }
+        if (!receiver.Item.Equips.Badge.TryGetValue(BadgeType.Buddy, out Item? receiverBadge)) {
+            return false;
+        }
+
+        return receiverBadge.Id != senderBadge.Id ||
```
kOchirasu, line 62: I'm a bit confused by all the checks here, they seem like the opposite of what we would want to check. Could you explain?

Zintixx: I could simplify it with just character Ids.
Yeah I think I had it flipped in reverse due to a character ID read. Pushed another update that should make more sense

PR #9, opened 2022-11-14
Mastery Level Rewards

Zintixx:


Maple2.Model/Metadata/Table/MasteryRewardTable.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+using Maple2.Model.Enum;
+
+namespace Maple2.Model.Metadata;
+
+public record MasteryRewardTable(IReadOnlyDictionary<MasteryType, MasteryRewardTable.Entry> Entries) : Table {
```
kOchirasu, line 6: this can be a nested dictionary `IReadOnlyDictionary<MasteryType, IReadOnlyDictionary<int, Entry>>`

There's an extension method I wrote that makes reading from these easier: https://github.com/kOchirasu/Maple2/blob/master/Maple2.Tools/Extensions/EnumerableExtensions.cs#L67

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
```diff
@@ -32,12 +32,52 @@ private enum Command : byte {
...
+            return;
+        }
+
+        if (!ItemMetadata.TryGet(levelMetadata.ItemId, out ItemMetadata? itemMetadata)) {
+            return;
+        }
+
+        var rewardItem = new Item(itemMetadata, levelMetadata.ItemRarity, levelMetadata.ItemAmount);
+
+        if (!session.Item.Inventory.CanAdd(rewardItem)) {
```
kOchirasu, line 69: Rather than checking `CanAdd`, you might as well just call `Add` which returns a bool. You would really only need to use `CanAdd` if there were other things you wanted to check after it, but before actually adding the item.
```cs
if (!session.Item.Inventory.Add(...)) {
    session.Send(ChatPacket.Alert(StringCode.s_err_inventory));
    return;
}

session.Player.Value.Unlock.MasteryRewardsClaimed.Add(rewardBoxDetails, true);
...
```

Maple2.Server.Game/Packets/UserEnvPacket.cs
```diff
@@ -40,4 +41,16 @@ private enum Command : byte {
 
         return pWriter;
     }
+
+    public static ByteWriter LoadClaimedRewards(IDictionary<int, bool> claimedRewards) {
+        var pWriter = Packet.Of(SendOp.UserEnv);
+        pWriter.Write<Command>(Command.MasteryRewardsClaimed);
+        pWriter.WriteInt(claimedRewards.Count);
+        foreach (KeyValuePair<int, bool> reward in claimedRewards) {
```
kOchirasu, line 49: I generally prefer decomposition since you can name the different values: `KeyValuePair<int, bool>` => `(int rewardId, bool isClaimed)`

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
```diff
@@ -32,12 +32,52 @@ private enum Command : byte {
...
                 break;
         }
     }
 
+    private void HandleReward(GameSession session, IByteReader packet) {
+        int rewardBoxDetails = packet.ReadInt();
+        var type = (MasteryType) (rewardBoxDetails / 1000);
+        int level = rewardBoxDetails % 100;
+
+        if (session.Player.Value.Unlock.MasteryRewardsClaimed.TryGetValue(rewardBoxDetails, out bool isClaimed) && isClaimed) {
```
kOchirasu, line 49: This check seems wrong, if the id is not in `MasteryRewardsClaimed`, shouldn't it be considered valid? Perhaps you instead mean: `if (!TryGetValue(...) || isClaimed)`?

Maple2.Server.Game/PacketHandlers/MasteryHandler.cs
```diff
@@ -32,12 +32,52 @@ private enum Command : byte {
...
+            session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_unknown));
+            return;
+        }
+
+        if (!TableMetadata.MasteryRewardTable.Entries.TryGetValue(type, out MasteryRewardTable.Entry? entry) || !entry.Levels.TryGetValue(level, out MasteryRewardTable.Level? levelMetadata)) {
```
kOchirasu, line 54: As mentioned above, once you switch to nested dictionaries, the extension method will allow you to do:
```cs
TableMetadata.MasteryRewardTable.Entries.TryGetValue(type, level, out ...);
```

PR #12, opened 2022-11-24
Fishing

Zintixx:


Maple2.Model/Metadata/Table/FishTable.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Maple2.Model.Metadata;
+
+public record FishTable(IReadOnlyDictionary<int, FishTable.Entry> Entries) : Table {
+    public record Entry(int Id,
+                        string FluidHabitat,
+                        IReadOnlyList<int> HabitatMapIds,
+                        int Mastery,
+                        short Rarity,
+                        int[] SmallSize,
```
kOchirasu, line 11: These arrays are always 2 values right? If that's true it might be good to define a new type:
```cs
...
FishTable.Range SmallSize,
FishTable.Range BigSize);

public record struct Range(int Min, int Max);
```

ref: https://github.com/kOchirasu/Maple2/blob/master/Maple2.Model/Metadata/Table/ItemOptionTable.cs#L35

Maple2.Model/Game/FishCatch.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Runtime.InteropServices;
+using Maple2.Model.Enum;
+using Maple2.Model.Metadata;
+
+namespace Maple2.Model.Game;
+
+[StructLayout(LayoutKind.Sequential, Pack = 4, Size = 12)]
+public struct FishCatch {
```
kOchirasu, line 8: Generally I only create readonly structs unless there's a good reason otherwise. Currently you are constructing `FishCatch` and then passing it to `AddMastery`. Instead you can pass `FishTable.Entry fishEntry, CaughtFishType type` to `AddMastery` and construct `FishCatch` there. That way you will not need to mutate this struct and can make it readonly.

---

Perhaps you can just eliminate this struct entirely though and pass the 4 values in it as parameters to `IncreaseMastery` since it's not used anywhere else.

Maple2.Database/Model/CharacterUnlock.cs
```diff
@@ -3,6 +3,7 @@
 using System.Linq;
 using Maple2.Database.Extensions;
 using Maple2.Model.Enum;
+using Maple2.Model.Game;
```
kOchirasu, line 6: Can we create a separate Database internal FishEntry type so we don't need Game import here. I try to keep a layer of separation between the two, hence all the conversion logic.

Maple2.File.Ingest/Mapper/TableMapper.cs
```diff
@@ -585,6 +592,61 @@ public class TableMapper : TypeMapper<TableMetadata> {
...
+
+    private FishTable ParseFish() {
+        var results = new Dictionary<int, FishTable.Entry>();
+
+        // Parse habitat and combine
+        var habitats = new Dictionary<int, IReadOnlyList<int>>();
+        foreach ((int id, FishHabitat habitat) in parser.ParseFishHabitat()) {
+            var habitatMaps = new List<int>();
+            foreach (string map in habitat.habitat) {
+                string mapString = map.Replace(",", "");
```
kOchirasu, line 612: This seems weird, are there extra commas or something?

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
```diff
@@ -0,0 +1,374 @@
...
+        Prepare = 0,
+        Stop = 1,
+        Catch = 8,
+        Start = 9,
+        FailMinigame = 10,
+    }
+
+    #region Autofac Autowired
+    // ReSharper disable MemberCanBePrivate.Global
+    public ItemMetadataStorage ItemMetadata { get; init; } = null!;
```
kOchirasu, line 33:
```cs
public required ItemMetadataStorage ItemMetadata { private get; init; }
public required TableMetadataStorage TableMetadata { private get; init; }
```

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
```diff
@@ -0,0 +1,374 @@
...
+    public TableMetadataStorage TableMetadata { private get; init; } = null!;
+    // ReSharper restore All
+    #endregion
+
+    public override void Handle(GameSession session, IByteReader packet) {
+        var function = packet.Read<Command>();
```
kOchirasu, line 39: nit: "function" => "command"

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
```diff
@@ -0,0 +1,374 @@
...
+            return;
+        }
+
+        if (!TableMetadata.FishingSpotTable.Entries.TryGetValue(session.Player.Value.Character.MapId, out FishingSpotTable.Entry? spotMetadata)) {
+            session.Send(FishingPacket.Error(FishingError.s_fishing_error_notexist_fish));
+            return;
+        }
+
+        if (session.Mastery[MasteryType.Fishing] < spotMetadata.MinMastery) {
+            session.Send(FishingPacket.Error(FishingError.s_fishing_error_lack_mastery));
```
kOchirasu, line 88: missing return after this error

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
```diff
@@ -0,0 +1,374 @@
...
+    private static IList<Vector3> GetFishingBlocks(Vector3 position, Vector3 rotation) {
+        /* Note this is all wrong and is a temp placeholder until the actual functionality is implemented.
+         Normally how it should work is we get the intercardinal direction that the player is facing closest to (Southwest, northwest, southeast, northeast)
+         We then scan in front of the player a 5x3x3 area for liquid blocks
+         The liquid block CANNOT have another block on top of it 
+         It cannot be ABOVE the block the player is standing on (i.e. scan the z axis same level as player and 2 blocks below max(?. May be more. Needs confirmation)
+         
+         For now, we will only be getting near-ish blocks. */
+
+        Vector3 roundedCoordinates = position;
```
kOchirasu, line 119: `position.Align()` already does this

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
```diff
@@ -0,0 +1,374 @@
...
+        
+        roundedCoordinates.X = (float) (Math.Round(roundedCoordinates.X / Constant.BlockSize) * Constant.BlockSize);
+        roundedCoordinates.Y = (float) (Math.Round(roundedCoordinates.Y / Constant.BlockSize) * Constant.BlockSize);
+        roundedCoordinates.Z = (float) (Math.Round(roundedCoordinates.Z / Constant.BlockSize) * Constant.BlockSize);
+
+        float direction = RotationDirectionUtil.GetClosestDirection(rotation);
+
+        var nearbyBlockCoordinates = new List<Vector3>();
```
kOchirasu, line 127: `VectorExtentions.BlocksInRange` might do what you want?

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
```diff
@@ -0,0 +1,374 @@
...
+    private void HandleCatch(GameSession session, IByteReader packet) {
+        bool success = packet.ReadBool();
+
+        if (session.GuideObject == null) {
+            return;
+        }
+
+        // TODO: Get the fishing block underneath the guide object and validate what fish can be caught there
+        // For now we will assume all fish in the map are valid
+        IEnumerable<FishTable.Entry> entries = TableMetadata.FishTable.Entries.Values.Where(x => x.HabitatMapIds.Contains(session.Player.Value.Character.MapId) && x.Mastery <= session.Mastery[MasteryType.Fishing]).ToArray();
```
kOchirasu, line 218: nit: "entry" instead of "x"

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
```diff
@@ -0,0 +1,374 @@
...
+                >= 0.85 and < 0.95 => 3,
+                >= 0.95 => 4,
+                _ => 1,
+            };
+
+            filteredFish = entries.Where(x => x.Rarity == rarity).ToList();
+        } while (filteredFish.Count == 0);
+
+        // get one fish from selected rarity
+        return filteredFish[Random.Shared.Next(0, filteredFish.Count)];
```
kOchirasu, line 343: I have an extension method for this: `filteredFish.Random()`

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
Zintixx: Not exactly. It needs to a directional range of blocks

Maple2.File.Ingest/Mapper/TableMapper.cs
Zintixx: There must be something with the parser on this one. It seems like it may be splitting the strings by space instead of a comma?

Maple2.Server.Game/Packets/FishingPacket.cs
```diff
@@ -0,0 +1,135 @@
...
+using Maple2.Model.Metadata;
+using Maple2.PacketLib.Tools;
+using Maple2.Server.Core.Constants;
+using Maple2.Server.Core.Packets;
+using Maple2.Tools.Extensions;
+
+namespace Maple2.Server.Game.Packets;
+
+public static class FishingPacket {
+    private enum Command : byte {
```
kOchirasu, line 16: There's actually 2 more ops: 10, 11 if you want to take a look
``` python
elif f == 10: # s_fishing_simulate
    count = add_int("count")
    for _ in range(count):
        add_int("unknown")
        add_int("unknown")
        count = add_int("unknown")
        for _ in range(count):
            add_int("unknown")
elif f == 11:
    b = add_bool("start")
    if b:
        pass # client sends fishing start (op 9)
```

I think op 11 might have something to do with autofishing? That bool is the same as the empty byte in op 8 and it might tell the client to keep fishing after catching?

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        if (session.Mastery[MasteryType.Fishing] < spotMetadata.MinMastery) {
+            session.Send(FishingPacket.Error(FishingError.s_fishing_error_lack_mastery));
+            return;
+        }
+
+        IList<Vector3> fishingBlocks = GetFishingBlocks(session.Player.Position, session.Player.Rotation);
+
+        session.GuideObject = session.Field?.SpawnGuideObject(session.Player, new FishingGuideObject(), GetGuideObjectPosition(fishingBlocks, session.Player.Position));
```
kOchirasu, line 94: `session.Field` is guaranteed to not be null here, so you don't need null coalescing, and therefore you also don't need the null ignore on `GuideObject` below

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
```diff
@@ -0,0 +1,377 @@
...
+    private static Vector3 GetGuideObjectPosition(IEnumerable<Vector3> fishingBlocks, Vector3 playerPosition) {
+        // get the fishing block closest to player
+        Vector3 guidePosition = fishingBlocks.MinBy(o => Math.Sqrt(Math.Pow(playerPosition.X - o.X, 2) + Math.Pow(playerPosition.Y - o.Y, 2)));
+
+        // Move guide 1 block up so it sits on top of the fishing block
+        guidePosition.Z += 150;
+        return guidePosition;
+    }
+
+    private static IList<Vector3> GetFishingBlocks(Vector3 position, Vector3 rotation) {
```
kOchirasu, line 111: rather than building up a list, you can just make this return `IEnumerable<Vector3>` and do `yield return checkBlock` for every block you would have added to the list.

For usage of this function you would do `GetFishingBlocks(position, rotation).ToArray()`

Maple2.Server.Game/PacketHandlers/FishingHandler.cs
```diff
@@ -0,0 +1,377 @@
...
+                }
+                break;
+        }
+
+        session.Mastery[MasteryType.Fishing] += exp;
+        session.Send(FishingPacket.IncreaseMastery(entry.Id, exp, fishType));
+    }
+
+    private static FishTable.Entry GetFishToCatch(IEnumerable<FishTable.Entry> entries) {
+
```
kOchirasu, line 329: nit: Not a fan of these newlines after function defs, can you remove them.

Maple2.Server.Game/Packets/FishingPacket.cs
Zintixx: Is this flipped around? 10 seems to be 11 and vice versa

kOchirasu: yea my bad, mixed them up

Maple2.Server.Game/Packets/FishingPacket.cs
```diff
@@ -0,0 +1,164 @@
...
+        return pWriter;
+    }
+
+    public static ByteWriter IncreaseMastery(int fishId, int exp, CaughtFishType fishType) {
+        var pWriter = Packet.Of(SendOp.Fishing);
+        pWriter.Write<Command>(Command.IncreaseMastery);
+        pWriter.WriteInt(fishId);
+        pWriter.WriteInt(exp);
+        pWriter.Write<CaughtFishType>(fishType);
+        pWriter.WriteShort(1); // fish grade/rarity ?
```
kOchirasu, line 61: it's not rarity, I think it's those like fishing tiers (beginner, intermediate, etc) maybe it does something special when you level up?

Maple2.Server.Game/Packets/FishingPacket.cs
```diff
@@ -0,0 +1,164 @@
...
+        return pWriter;
+    }
+
+    public static ByteWriter CatchFish(int id, int size, FishEntry? fish = null) {
+        var pWriter = Packet.Of(SendOp.Fishing);
+        pWriter.Write<Command>(Command.CatchFish);
+        pWriter.WriteInt(id);
+        pWriter.WriteInt(size);
+        pWriter.WriteBool(fish != null);
+        pWriter.WriteByte();
```
kOchirasu, line 121: this is also `autoFish`

PR #13, opened 2022-11-30
Buddy Emote Fix

Zintixx:
Fixes:
Correct opcode for Buddy Emote
Start packet


kOchirasu:
so that's why it didn't work... I just gave up

PR #15, opened 2022-12-04
Character Name Change Coupon

Zintixx:


Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -261,4 +264,21 @@ public class ItemUseHandler : PacketHandler<GameSession> {
             session.Send(ItemUsePacket.CharacterSlotAdded());
         }
     }
+
+    private void HandleChangeCharacterName(GameSession session, IByteReader packet, Item item) {
+        string newName = packet.ReadUnicodeString();
```
kOchirasu, line 269: Name needs to be validated here as well, a bad actor could just not call the function to check names.

PR #16, opened 2023-01-04
Initial Npc Talking handling of TalkType Flags

Zintixx:


Maple2.Server.Game/Scripting/Npc/NpcScript.cs
```diff
@@ -13,6 +13,7 @@ public class NpcScript {
     }
 
     public bool Begin() {
+        context.SetTalkTypeFlags(script.first());
```
kOchirasu, line 16: Can we move this into the constructor so TalkType can remain `readonly`?

Maple2.Server.Game/Scripting/Npc/NpcScriptContext.cs
```diff
@@ -98,6 +85,37 @@ public class NpcScriptContext {
         return true;
     }
 
+    public void SetTalkTypeFlags(int firstState) {
```
kOchirasu, line 88: I feel like this function can be reorganized to better accommodate quests in the future:

```cs
switch (Npc.Value.Metadata.Basic.Kind) {
    int options = 0;
    case 1 or > 10 and < 20:
        TalkType |= NpcTalkType.Dialog;
        options++;
        break;
    case ...
        TalkType |= NpcTalkType.Dialog
}

if (Metadata.States.TryGetValue(firstState, out ScriptState? scriptState)) {
    if (scriptState.Type == ScriptStateType.Job) {
        TalkType |= NpcTalkType.Dialog;
    } else {
        TalkType |= NpcTalkType.Talk;
        options++;
    }
}

if (options > 1) {
    TalkType |= NpcTalkType.Select;
}
````

Maple2.Server.Game/Scripting/Npc/NpcScript.cs
kOchirasu: can't make it `readonly` anyways, but I think it's still worth moving to the constructor.

Maple2.Server.Game/Scripting/Npc/NpcScriptContext.cs
kOchirasu: Not sure if the logic is 100% accurate. Is a Job script counted separate from first script? (e.g. can an Npc have both and does that require Select?)

PR #17, opened 2023-01-07
Item Expiration

Zintixx:


Maple2.Model/Game/Item/Item.cs
```diff
@@ -173,6 +176,28 @@ public class Item : IByteSerializable, IByteDeserializable {
...
+
+        if (Metadata.Life.ExpirationPeriod != default) {
+            expirationTime = (long) (Metadata.Life.ExpirationPeriod.ToUniversalTime() - DateTime.UnixEpoch).TotalSeconds;
+        } else if (Metadata.Life.UsePeriod > 0) {
+            expirationTime = (long) (DateTime.Now.ToUniversalTime() - DateTime.UnixEpoch).TotalSeconds + Metadata.Life.UsePeriod;
+        } else if (Metadata.Life.ExpirationType != ItemExpirationType.None) {
+            expirationTime = Metadata.Life.ExpirationType switch
+            {
+                ItemExpirationType.Months => Metadata.Life.ExpirationTypeDuration * 2628000 + (long) (DateTime.Now.ToUniversalTime() - DateTime.UnixEpoch).TotalSeconds,
+                ItemExpirationType.Weeks => Metadata.Life.ExpirationTypeDuration * 604800 + (long) (DateTime.Now.ToUniversalTime() - DateTime.UnixEpoch).TotalSeconds,
```
kOchirasu, line 193: `TimeSpan.FromDays(...).Seconds`

Maple2.Model/Game/Item/Item.cs
```diff
@@ -173,6 +176,28 @@ public class Item : IByteSerializable, IByteDeserializable {
...
+        }
+
+        if (Metadata.Life.ExpirationPeriod != default) {
+            expirationTime = (long) (Metadata.Life.ExpirationPeriod.ToUniversalTime() - DateTime.UnixEpoch).TotalSeconds;
+        } else if (Metadata.Life.UsePeriod > 0) {
+            expirationTime = (long) (DateTime.Now.ToUniversalTime() - DateTime.UnixEpoch).TotalSeconds + Metadata.Life.UsePeriod;
+        } else if (Metadata.Life.ExpirationType != ItemExpirationType.None) {
+            expirationTime = Metadata.Life.ExpirationType switch
+            {
+                ItemExpirationType.Months => Metadata.Life.ExpirationTypeDuration * 2628000 + (long) (DateTime.Now.ToUniversalTime() - DateTime.UnixEpoch).TotalSeconds,
```
kOchirasu, line 192: How was this seconds/month (2628000) value determined?

Maple2.File.Ingest/Mapper/ItemMapper.cs
```diff
@@ -82,6 +88,12 @@ public class ItemMapper : TypeMapper<ItemMetadata> {
                     .Select(slot => Enum.Parse<EquipSlot>(slot.name, true))
                     .ToArray(),
                 Mesh: data.ucc.mesh,
+                Life: new ItemMetadataLife(
+                    UsePeriod: data.life.usePeriod,
+                    ExpirationPeriod: expirationTime,
+                    ExpirationType: (ItemExpirationType) data.life.expirationType,
```
kOchirasu, line 94: I feel like we can just do the conversion here at parse time and just store duration in seconds rather than calculating it in `Item.cs`

Basically just store `ExpirationDuration` rather than `ExpirationType`+`ExpirationTypeDuration`

Maple2.Model/Metadata/ItemMetadata.cs
```diff
@@ -15,6 +17,12 @@ public record ItemMetadata(
     ItemMetadataMusic? Music,
     ItemMetadataHousing? Housing) : ISearchResult;
 
+public record ItemMetadataLife(
+    int UsePeriod,
+    DateTime ExpirationPeriod,
```
kOchirasu, line 22: It's probably better not to store as `DateTime` for metadata entries since we don't have a conversion layer. (Metadata types are directly used by both DB and Server due to immutability). MySQL `DateTime` is stored using local system timezone which can technically vary, while I've standardized the game layer to use UTC (even though it's technically wrong, at least it's consistent)

tl;dr Let's use long UTC epoch timestamp for now.

Maple2.File.Ingest/Mapper/ItemMapper.cs
```diff
@@ -82,6 +101,10 @@ public class ItemMapper : TypeMapper<ItemMetadata> {
                     .Select(slot => Enum.Parse<EquipSlot>(slot.name, true))
                     .ToArray(),
                 Mesh: data.ucc.mesh,
+                Life: new ItemMetadataLife(
```
kOchirasu, line 104: `Life` is marked nullable, but it's never null here

PR #18, opened 2023-01-11
Item Exchange Scrolls

Zintixx:


Maple2.Model/Metadata/Table/ScrollTable.cs
```diff
@@ -16,7 +16,14 @@ public record EnchantScrollMetadata(
...
 
-public record ItemExchangeScrollMetadata();
+public record ItemExchangeScrollMetadata(
+    Ingredient RecipeScroll,
+    Ingredient RewardItem,
+    int TradeCountDeduction,
+    long RequiredMeso,
+    IReadOnlyList<Ingredient> RequiredItems);
+
+public record Ingredient(int ItemId, short Rarity, int Amount, ItemTag Tag);
```
kOchirasu, line 26: This is too common of a class name to have in the top-level namespace. If you want to keep this you should move it somewhere common and use it everywhere. (There are definitely other places where we need this struct) Otherwise just nest it in `ItemExchangeScrollMetadata`.

Also Rarity is commonly used as an `int`, but if you mostly need it as a `short` here that's fine.

Maple2.Server.Game/PacketHandlers/ItemExchangeScrollHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+        
+        lock (session.Item) {
+            // Check if player has enough scrolls
+            List<Item> scrolls = session.Item.Inventory.Find(scrollMetadata.RecipeScroll.ItemId, scrollMetadata.RecipeScroll.Rarity).ToList();
+            if (scrolls.Sum(x => x.Amount) < scrollMetadata.RecipeScroll.Amount * quantity) {
+                session.Send(ItemExchangeScrollPacket.Error(ItemExchangeScrollError.s_itemslot_exchange_count_invalid));
+                return;
+            }
+            
+            // Check for ingredients
```
kOchirasu, line 70: I don't like that we keep duplicating this logic. Can you add a helper like: https://github.com/kOchirasu/Maple2/blob/master/Maple2.Server.Game/Manager/Items/InventoryManager.cs#L222

Replace the usage here as well: https://github.com/kOchirasu/Maple2/blob/master/Maple2.Server.Game/PacketHandlers/MasteryHandler.cs#L99-L163

Maple2.Server.Game/Packets/ItemExchangeScrollPacket.cs
```diff
@@ -0,0 +1,37 @@
+﻿using Maple2.Model.Error;
+using Maple2.PacketLib.Tools;
+using Maple2.Server.Core.Constants;
+using Maple2.Server.Core.Packets;
+
+namespace Maple2.Server.Game.Packets;
+
+public static class ItemExchangeScrollPacket {
+    private enum Command : byte {
+        Mode0 = 0,
```
kOchirasu, line 10: `Unknown0` for consistency

Maple2.Server.Game/PacketHandlers/ItemExchangeScrollHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+        
+        session.Currency.Meso -= scrollMetadata.RequiredMeso;
+
+        if (!ItemMetadata.TryGet(scrollMetadata.RewardItem.ItemId, out ItemMetadata? itemMetadata)) {
+            return;
+        }
+            
+        var rewardItem = new Item(itemMetadata, scrollMetadata.RewardItem.Rarity, scrollMetadata.RewardItem.Amount * quantity);
+        if (!session.Item.Inventory.Add(rewardItem, true)) {
+            // TODO: mail player
```
kOchirasu, line 155: there should be a return here and/or some error otherwise it will be treated as success

Zintixx: That's intentional. It's still a success, being mailed to the player is the only difference.

kOchirasu: Since it's not currently being mailed I felt like a failure would make more sense but it's fine.

PR #21, opened 2023-02-20
Game Events - part 2 - Blue Marble/Mapleopoly & GameEventUserValues

Zintixx:
I figured that GameEventUserValues should have their own table in order to compliment with modifying events on the fly.


Maple2.Server.Game/PacketHandlers/MapleopolyHandler.cs
```diff
@@ -0,0 +1,183 @@
...
+        GameEvent? gameEvent = db.FindEvent(nameof(BlueMarble));
+        var blueMarble = gameEvent?.EventInfo as BlueMarble;
+        if (blueMarble == null || gameEvent == null) {
+            // TODO: Find error to state event is not active
+            return;
+        }
+
+        blueMarble.Tiles = blueMarble.Tiles.OrderBy(tile => tile.Position).ToList();
+
+        GameEventUserValue freeRollValue = session.Config.GetGameEventUserValue(GameEventUserValueType.MapleopolyFreeRollAmount, gameEvent);
```
kOchirasu, line 45: might be worth it to just `int.Parse` these up front since it's always done anyways.

Maple2.Server.Game/PacketHandlers/MapleopolyHandler.cs
```diff
@@ -0,0 +1,183 @@
...
+    #endregion
+
+    private enum Command : byte {
+        Load = 0,
+        Roll = 1,
+        Result = 3,
+    }
+
+    public override void Handle(GameSession session, IByteReader packet) {
+        using GameStorage.Request db = GameStorage.Context();
```
kOchirasu, line 35: I would move this all into a helper function and call it in the respective Handle functions. This way we don't run this code for invalid `Command`.

Maple2.Server.Game/PacketHandlers/MapleopolyHandler.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        // roll two dice
+        int dice1 = Random.Shared.Next(1, 7);
+        int dice2 = Random.Shared.Next(1, 7);
+
+        int.TryParse(totalTileValue.Value, out int totalTiles);
+        totalTiles += dice1 + dice2;
+        if (dice1 == dice2) {
+            freeRolls++;
```
kOchirasu, line 95: this updated value is never saved

Maple2.Server.Game/Packets/MapleopolyPacket.cs
```diff
@@ -0,0 +1,68 @@
...
+        var pWriter = Packet.Of(SendOp.Mapleopoly);
+        pWriter.Write<Command>(Command.Load);
+        pWriter.WriteInt(totalTileCount);
+        pWriter.WriteInt(freeRollAmount);
+        pWriter.WriteInt(Constant.MapleopolyTicketItemId);
+        pWriter.WriteInt(playerTicketAmount);
+        pWriter.WriteInt(tiles.Count);
+
+        foreach (BlueMarbleTile tile in tiles) {
+            pWriter.WriteClass(tile);
```
kOchirasu, line 31: nit: `WriteClass<BlueMarbleTile>(tile);` for safety.

Maple2.Server.Game/PacketHandlers/MapleopolyHandler.cs
```diff
@@ -0,0 +1,191 @@
...
+                HandleRoll(session);
+                return;
+            case Command.Result:
+                HandleResult(session);
+                return;
+        }
+    }
+
+    private void HandleLoad(GameSession session) {
+        using GameStorage.Request db = GameStorage.Context();
```
kOchirasu, line 50: Is it necessary to read the event from db every time? I assume the same event is used across all players on the server. It might make more sense to manage this centrally (e.g. `GameServer`, and retrieve it from memory)

Rough idea of what this would look like:

`GameServer.cs`
```cs
public T? FindEvent<T>() where T : GameEvent {
    if (!eventCache.TryGetValue(typeof(T).Name, out GameEvent? gameEvent) {
        gameEvent = ...; // Lookup from db
        eventCache[typeof(T).Name] = gameEvent;
    }

    // You can check expiry here as well if you want.
    return gameEvent as T;
}
```

`GameSession.cs`
```cs
public T? FindEvent<T>() where T : GameEvent {
    return server.FindEvent<T>();
}
```

`MapleopolyHandler.cs`
```cs
BlueMarble? blueMarble = session.FindEvent<BlueMarble>();
if (blueMarble == null {
    return;
}
...
```

Maple2.Server.Game/PacketHandlers/MapleopolyHandler.cs
```diff
@@ -0,0 +1,191 @@
...
+        GameEvent? gameEvent = db.FindEvent(nameof(BlueMarble));
+        if (gameEvent?.EventInfo is not BlueMarble blueMarble) {
+            // TODO: Find error to state event is not active
+            return;
+        }
+
+        blueMarble.Tiles = blueMarble.Tiles.OrderBy(tile => tile.Position).ToList();
+
+        int.TryParse(session.GameEventUserValue.Get(GameEventUserValueType.MapleopolyFreeRollAmount, gameEvent).Value, out int freeRollValue);
```
kOchirasu, line 59: This is kinda ugly, what do you think about adding `.Int()`/`.String()` functions on `GameEventUserValueManager`. Open to other ideas too. I also don't like the fact that you need to pass `gameEvent` just for the scenario that the value needs to be instantiated, but can't think of a better structure right now.

Example:
```cs
int freeRollValue = session.GameEventUserValue.Get(GameEventUserValueType.MapleopolyFreeRollAmount, gameEvent).Int();
```

Maple2.Server.Game/PacketHandlers/MapleopolyHandler.cs
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        GameEventUserValue totalTrips = session.GameEventUserValue.Get(GameEventUserValueType.MapleopolyTotalTrips, gameEvent);
+        int.TryParse(totalTrips.Value, out int trips);
+        int totalNewTrips = totalTileValue / blueMarble.Tiles.Count;
+
+        if (totalNewTrips > trips) {
+            int difference = totalNewTrips - trips;
+
+            for (int tripCount = 0; tripCount < difference; tripCount++) {
```
kOchirasu, line 163: What about:
```cs
while (trips < totalNewTrips) {
```

Maple2.Server.Game/PacketHandlers/MapleopolyHandler.cs
```diff
@@ -0,0 +1,191 @@
...
+                    var trip0Item = new Item(trip0Metadata, entry1.Item.ItemRarity, entry1.Item.ItemAmount);
+                    if (!session.Item.Inventory.Add(trip0Item, true)) {
+                        // TODO: mail to player
+                    }
+                }
+
+                // Check if there's any other item to give for hitting a specific number of trips
+                var entry2 = blueMarble.Entries.FirstOrDefault(entry => entry.TripAmount == trips);
+                if (entry2 == default || !ItemMetadata.TryGet(entry2.Item.ItemId, out ItemMetadata? tripMetadata)) {
+                    continue;
```
kOchirasu, line 178: i think it's more clear to follow the same logic pattern as above and not use `continue` here.

Maple2.Server.Game/PacketHandlers/MapleopolyHandler.cs
```diff
@@ -0,0 +1,191 @@
...
+                break;
+            case BlueMarbleTileType.Lose:
+            case BlueMarbleTileType.RollAgain:
+            case BlueMarbleTileType.Trap:
+            default:
+                Logger.Warning($"Unhandled Mapleopoly tile type: {tile.Type}");
+                break;
+        }
+
+        GameEventUserValue totalTrips = session.GameEventUserValue.Get(GameEventUserValueType.MapleopolyTotalTrips, gameEvent);
```
kOchirasu, line 156: It might help to split this logic into a helper function:

Helper:
```cs
private void RewardNewTrips(GameSession session, GameEvent gameEvent, int totalNewTrips) {
    GameEventUserValue totalTrips = session.GameEventUserValue.Get(GameEventUserValueType.MapleopolyTotalTrips, gameEvent);
    int.TryParse(totalTrips.Value, out int trips);
    while (trips < totalNewTrips) {
        ...
    }
}
```

Usage:
```cs
int totalNewTrips = totalTileValue / blueMarble.Tiles.Count;
RewardNewTrips(session, gameEvent, totalNewTrips);
session.GameEventUserValue[GameEventUserValueType.MapleopolyTotalTrips] = totalNewTrips;
...
```

Maple2.Server.Game/PacketHandlers/MapleopolyHandler.cs
Zintixx: I see what you mean. I would need to change `FindEvent` (in both GameServer and GameSession) to something of the following:

```csharp    
public GameEvent? FindEvent<T>() where T : GameEventInfo {
...
        return gameEvent;
    }
```

considering I need to search by GameEventInfo but get GameEvent returned instead

However would it make sense to null suppress in MapleopolyHandler?

```csharp
        GameEvent? gameEvent = session.FindEvent<BlueMarble>();
        if (gameEvent == null) {
            return;
        }
        BlueMarble blueMarble = (gameEvent.EventInfo as BlueMarble)!;
        ```

Considering this was already checked if it was BlueMarble with both FindEvent functions, it should be safe to assume it _will_ be BlueMarble, no?

Comment (kOchirasu) [reply Maple2.Server.Game/PacketHandlers/MapleopolyHandler.cs:50]: Oh right, didn't realize it was `GameEvent` vs `GameEventInfo`. I think what you have is reasonable, I had an idea around making `GameEvent` into `GameEvent<T> where T : GameEventInfo`, but it gets complicated and I'm not sure if it even works.

To access db, just add `GameStorage gameStorage` to the constructor and it will be injected by autofac.

File: Maple2.Server.Game/GameServer.cs [hunk 10]
```diff
@@ -18,14 +19,22 @@ public class GameServer : Server<GameSession> {
     private readonly FieldManager.Factory fieldFactory;
     private readonly HashSet<GameSession> connectingSessions;
     private readonly Dictionary<long, GameSession> sessions;
+    private readonly Dictionary<string, GameEvent> eventCache = new();
 
     public int Channel => Target.GameChannel;
+    
+    #region Autofac Autowired
```
kOchirasu, line 26: You can't do this unless you have `PropertiesAutowired()` which is why you have `context.Resolve<GameStorage>()`. Just define this normally to avoid confusion:

```cs
private GameStorage gameStorage;
```

Also you shouldn't need to use `context.Resolve<GameStorage>()`, just add `GameStorage gameStorage` as a parameter to the constructor.

PR #24, opened 2023-02-25
Super Chat Handler

Zintixx:


Maple2.Server.Game/PacketHandlers/SuperChatHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+            case Command.Deselect:
+                HandleDeselect(session);
+                return;
+        }
+    }
+
+    private static void HandleSelect(GameSession session, IByteReader packet) {
+        int itemId = packet.ReadInt();
+
+        Item? superChatItem = session.Item.Inventory.Find(itemId).First();
```
kOchirasu, line 35: Should be checking `Function.Type` here (`SuperWorldChat`)

Maple2.Server.Game/PacketHandlers/UserChatHandler.cs
```diff
@@ -179,17 +179,35 @@ public class UserChatHandler : PacketHandler<GameSession> {
...
+            Super = new ChatRequest.Types.Super {ItemId = superChatItem.Id},
         };
 
-        try {
-            World.Chat(request);
-        } catch (RpcException) { }
+        if (session.Item.Inventory.Consume(superChatItem.Uid, 1)) {
+            try {
+                session.Send(SuperChatPacket.Deselect(session.Player.ObjectId));
+                session.Player.Value.Character.SuperChatId = 0;
```
kOchirasu, line 206: nit: set `SuperChatId = 0` before sending `Deselect` packet.

Maple2.Server.Game/PacketHandlers/SuperChatHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+        if (!int.TryParse(superChatItem.Metadata.Function?.Parameters.Split(",").First(), out int superChatId)) {
+            return;
+        }
+
+        session.Player.Value.Character.SuperChatId = superChatId;
```
kOchirasu, line 40: I feel like we should just store the `itemId` which makes it easier to consume the item when it's used.

Maple2.Server.Game/PacketHandlers/UserChatHandler.cs
```diff
@@ -179,17 +179,35 @@ public class UserChatHandler : PacketHandler<GameSession> {
...
+            return;
+        }
+
         var request = new ChatRequest {
             AccountId = session.AccountId,
             CharacterId = session.CharacterId,
             Name = session.PlayerName,
             Message = message,
-            Super = new ChatRequest.Types.Super {ItemId = 0},
+            Super = new ChatRequest.Types.Super {ItemId = superChatItem.Id},
```
kOchirasu, line 200: I think this value is more likely to be the id from `Function` even though they are probably the same.

Maple2.Server.Game/PacketHandlers/SuperChatHandler.cs
Zintixx: so like save both superChatId and ItemId?

kOchirasu: you don't need to save `superChatId` because with `itemId` you can get `ItemMetadata` which gives you the same `superChatId` again

Zintixx: But we do need to save it. it goes in the FieldAddUser packet. this is so if a user changes fields, their superchat is still enabled/selected

kOchirasu: Oh I see, it's fine to save both then if we need it

PR #26, opened 2023-02-28
Game Events - part 3 - Attendance Event

Zintixx:
Also includes the fix on GameEventUserValue DB key. Applying a migration to change the primary key due to constraint so wasn't sure if there was a better solution other than altering the table or the previous migration


Maple2.Model/Game/Event/AttendGift.cs
```diff
@@ -0,0 +1,70 @@
...
+            writer.WriteInt();
+        }
+
+        writer.WriteInt(Days.Count);
+        foreach (AttendanceDayItem day in Days.OrderBy(day => day.Day)) {
+            writer.WriteClass<AttendanceDayItem>(day);
+        }
+    }
+    
+    public class AttendanceDayItem : IByteSerializable{
```
kOchirasu, line 54: nit: space before '{'

Maple2.Server.Game/Packets/AttendancePacket.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Maple2.Model.Error;
+using Maple2.PacketLib.Tools;
+using Maple2.Server.Core.Constants;
+using Maple2.Server.Core.Packets;
+
+namespace Maple2.Server.Game.Packets;
+
+public static class AttendancePacket {
```
kOchirasu, line 8: other structs:
```py
if f == 5:
    pass
elif f == 6:
    count = add_int("count")
    for i in range(count):
        add_int("Id")
        add_long("Timestamp")
elif f == 7:
    add_int("unknown")
elif f == 9:
    notice = add_byte("NoticeId")
    if notice == 1:
        pass # s_attendGift_payAttend_result_lackMoney
    elif notice == 2:
        pass # s_attendGift_payAttend_result_lackMerat
    elif notice == 3:
        pass # s_attendGift_payAttend_result_hasNotCoupon
    elif notice == 5:
        pass # s_attendGift_item_attend_already_used
    elif notice == 6:
        pass # s_attendGift_item_attend_not_found_evnet
```

Maple2.Server.Game/PacketHandlers/AttendanceHandler.cs
```diff
@@ -0,0 +1,162 @@
...
+public class AttendanceHandler : PacketHandler<GameSession> {
+    public override RecvOp OpCode => RecvOp.Attendance;
+
+    #region Autofac Autowired
+    // ReSharper disable MemberCanBePrivate.Global
+    public required ItemMetadataStorage ItemMetadata { private get; init; }
+    // ReSharper restore All
+    #endregion
+
+    private enum Command : byte {
```
kOchirasu, line 27: all ops:

```py
f = add_byte("function")
if f == 0:
    pass
elif f == 2:
    pass
elif f == 3:
    pass
elif f == 4:
    add_int("unknown")
    add_long("unknown")
elif f == 8:
    add_int("unknown")
```

kOchirasu: I didn't see 1 anywhere, but maybe I missed it

Maple2.Model/Game/Event/AttendGift.cs
kOchirasu, line 54: Was this intentionally a nested class? If so, maybe you can use a name that's less redundant since it will always be scoped with `AttendGift.`

Maple2.Database/Model/Event/GameEvent.cs
```diff
@@ -99,10 +102,24 @@ internal abstract class GameEventInfo {
                 Entries = blueMarble.Entries,
                 Tiles = blueMarble.Tiles,
             },
+            Maple2.Model.Game.Event.AttendGift attendGift => new AttendGift {
```
kOchirasu, line 105: I don't think we need any of the conversions in this file because it's actually duplicating the existing implicit constructors. You should be able to invoke the conversion by doing  a cast similar to:

https://github.com/kOchirasu/Maple2/blob/d98fead758a0d0eea2ef4ea6d29032f80fdb1e26/Maple2.Database/Model/Item/Item.cs#L61-L65

PR #27, opened 2023-03-19
Item Box

Zintixx:


Maple2.File.Ingest/Mapper/TableMapper.cs
```diff
@@ -1008,4 +1010,69 @@ public class TableMapper : TypeMapper<TableMetadata> {
...
+        results = GetIndividualItemDropTable(results, parser.ParseIndividualItemGearBox());
+        results = GetIndividualItemDropTable(results, parser.ParseIndividualItemDropEventNpc());
+        results = GetIndividualItemDropTable(results, parser.ParseIndividualItemDropNewGacha());
+        results = GetIndividualItemDropTable(results, parser.ParseIndividualItemDropQuestMob());
+        results = GetIndividualItemDropTable(results, parser.ParseIndividualItemDropQuestObj());
+
+        return new IndividualItemDropTable(results);
+    }
+    
+    private Dictionary<int, Dictionary<byte, IList<IndividualItemDropTable.Entry>>> GetIndividualItemDropTable(Dictionary<int, Dictionary<byte, IList<IndividualItemDropTable.Entry>>> results, IEnumerable<(int Id, IDictionary<byte,List<IndividualItemDrop>>)> parser) {
```
kOchirasu, line 1030: nit: MergeIndividualItemDropTable

Maple2.Model/Metadata/Table/IndividualItemDropTable.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
+using Maple2.Model.Enum;
+
+namespace Maple2.Model.Metadata;
+
+public record IndividualItemDropTable(
+    Dictionary<int, Dictionary<byte, IList<IndividualItemDropTable.Entry>>> Entries): Table {
+
+    public record Entry(
+        IList<int> ItemIds,
```
kOchirasu, line 10: Use Array since it's immutable

Maple2.File.Ingest/Mapper/TableMapper.cs
```diff
@@ -1008,4 +1010,69 @@ public class TableMapper : TypeMapper<TableMetadata> {
...
+                        SmartDropRate: drop.smartDropRate,
+                        Rarity: drop.PackageUIShowGrade,
+                        EnchantLevel: drop.enchantLevel,
+                        ReduceTradeCount: drop.tradableCountDeduction,
+                        ReduceRepackLimit: drop.rePackingLimitCountDeduction,
+                        Bind: drop.isBindCharacter,
+                        MinCount: (int) minCount,
+                        MaxCount: (int) maxCount);
+
+                    if (!results.ContainsKey(id)) {
```
kOchirasu, line 1060: Seems like this might miss the case where `id` is contained, but `dropGroup` is not contained? Or maybe this never happens.

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -320,4 +327,16 @@ public class ItemUseHandler : PacketHandler<GameSession> {
 
         session.Config.UpdatePremiumTime(hours);
     }
+    
+    private static void HandleSelectItemBox(GameSession session, IByteReader packet, Item item) {
+        short unknown = packet.ReadShort();
+        int index = packet.ReadShort() - 48;
```
kOchirasu, line 333: what does 48 mean here? should it be a named constant?

Maple2.Server.Game/Packets/ItemBoxPacket.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Maple2.Model.Error;
+using Maple2.PacketLib.Tools;
+using Maple2.Server.Core.Constants;
+using Maple2.Server.Core.Packets;
+
+namespace Maple2.Server.Game.Packets;
+
+public static class ItemBoxPacket {
+
+    public static ByteWriter Open(int sourceItemId, int amount, ItemBoxError error) {
```
kOchirasu, line 10: Can we split this into 2 functions, when that is never error and one that is always error? Or maybe have a default value for error so you don't need to pass it in.

Maple2.Server.Game/Manager/ItemBoxManager.cs
```diff
@@ -0,0 +1,284 @@
...
+using Maple2.Model.Enum;
+using Maple2.Model.Error;
+using Maple2.Model.Game;
+using Maple2.Model.Metadata;
+using Maple2.Server.Game.Session;
+
+namespace Maple2.Server.Game.Manager;
+
+public class ItemBoxManager {
+    #region Autofac Autowired
```
kOchirasu, line 15: don't need to pass these in, you can already access them via `session.ItemMetadata` and `session.TableMetadata`

Maple2.Server.Game/Manager/Items/InventoryManager.cs
```diff
@@ -12,6 +12,7 @@
 using Maple2.Server.Game.Packets;
 using Maple2.Server.Game.Session;
 using Maple2.Tools.Extensions;
+using Microsoft.Scripting.Utils;
```
kOchirasu, line 15: remove

Maple2.Server.Game/Manager/Items/InventoryManager.cs
```diff
@@ -271,11 +272,15 @@ public class InventoryManager {
                 ingredient => ingredient.ItemId,
                 ingredient => session.Item.Inventory.Find(ingredient.ItemId, ingredient.Rarity).ToList()
             );
-            Dictionary<ItemTag, IList<Item>> materialsByTag = components.ToDictionary(
-                ingredient => ingredient.Tag,
-                ingredient => session.Item.Inventory.Filter(item => item.Metadata.Property.Tag == ingredient.Tag)
-            );
-            
+            var materialsByTag = new Dictionary<ItemTag, IList<Item>>();
+            foreach(ItemComponent ingredient in components) {
```
kOchirasu, line 276: nit: space after `foreach`

Maple2.Server.Game/Manager/ItemBoxManager.cs
```diff
@@ -0,0 +1,284 @@
...
+        return item.Metadata.Function?.Type switch {
+            ItemFunction.SelectItemBox => SelectItemBox(item, itemBoxParams, index, count),
+            ItemFunction.OpenItemBox => OpenItemBox(item, itemBoxParams, count),
+            ItemFunction.OpenItemBoxWithKey => OpenItemBoxWithKey(item, itemBoxParams, count),
+            _ => throw new ArgumentOutOfRangeException(item.Metadata.Function?.Type.ToString(), "Invalid box type"),
+        };
+    }
+
+    private ItemBoxError SelectItemBox(Item item, int[] itemBoxParams, int index, int count = 1) {
+        var error = ItemBoxError.ok;
```
kOchirasu, line 45: define variables where they will be used, in this case around L68

Maple2.Server.Game/Manager/ItemBoxManager.cs
```diff
@@ -0,0 +1,284 @@
...
+        // SelectItemBox disregards DropGroup
+        IList<IndividualItemDropTable.Entry> drops = session.ItemBox.FilterDrops(dropGroupTable.Values.SelectMany(x => x).ToList());
+
+        if (drops.Count == 0) {
+            return ItemBoxError.s_err_cannot_open_multi_itembox_inventory_fail;
+        }
+
+        IndividualItemDropTable.Entry selectedEntry = drops[index];
+        for (int startCount = 0; startCount < count; startCount++) {
+            if (!session.Item.Inventory.ConsumeItemComponents(new List<ItemComponent>() {
```
kOchirasu, line 70: Here and elsewhere, prefer array instead of List if you are don't need to add to it afterwards.

```cs
new []{new ItemComponent(item.Id, -1, itemBoxParams[0], ItemTag.None)}
```

Maple2.Server.Game/Manager/ItemBoxManager.cs
```diff
@@ -0,0 +1,284 @@
...
+            }
+            BoxCount++;
+            if (error != ItemBoxError.ok) {
+                return error;
+            }
+        }
+        return error;
+    }
+
+    private ItemBoxError OpenItemBox(Item item, int[] itemBoxParams, int count = 1) {
```
kOchirasu, line 91: rather than passing `int[] itemBoxParams` can you split it into individual parameters? I think it makes the code easier to follow since you can name the parameters.

```cs
private ItemBoxError OpenItemBox(Item item, int param1, int param2, ..., int count = 1)
```

Maple2.Server.Game/Manager/ItemBoxManager.cs
```diff
@@ -0,0 +1,284 @@
...
+public class ItemBoxManager {
+    #region Autofac Autowired
+    // ReSharper disable MemberCanBePrivate.Global
+    public required ItemMetadataStorage ItemMetadata { private get; init; }
+    public required TableMetadataStorage TableMetadata { private get; init; }
+    // ReSharper restore All
+    #endregion
+
+    private readonly GameSession session;
+    public int BoxCount;
```
kOchirasu, line 23: `{get; private set;}`?

Maple2.Server.Game/Manager/Items/InventoryManager.cs
Zintixx: This is needed to add range on line 281

Maple2.Server.Game/Packets/ItemBoxPacket.cs
Zintixx: As far as I know, we would always have to pass it in.

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
Zintixx: I have no idea. subtracting 48 gets the correct index. Where does this magic number come from? who knows.

Maple2.File.Ingest/Mapper/TableMapper.cs
Zintixx: That is handled a few lines below with the else if

Maple2.File.Ingest/Mapper/TableMapper.cs
```diff
@@ -1008,4 +1010,69 @@ public class TableMapper : TypeMapper<TableMetadata> {
...
+        results = MergeIndividualItemDropTable(results, parser.ParseIndividualItemDropEventNpc());
+        results = MergeIndividualItemDropTable(results, parser.ParseIndividualItemDropNewGacha());
+        results = MergeIndividualItemDropTable(results, parser.ParseIndividualItemDropQuestMob());
+        results = MergeIndividualItemDropTable(results, parser.ParseIndividualItemDropQuestObj());
+
+        return new IndividualItemDropTable(results);
+    }
+    
+    private Dictionary<int, Dictionary<byte, IList<IndividualItemDropTable.Entry>>> MergeIndividualItemDropTable(Dictionary<int, Dictionary<byte, IList<IndividualItemDropTable.Entry>>> results, IEnumerable<(int Id, IDictionary<byte,List<IndividualItemDrop>>)> parser) {
+        foreach ((int id,  IDictionary<byte,List<IndividualItemDrop>> dict) in parser) {
```
kOchirasu, line 1031: nit: 2 spaces before IDictionary

Maple2.File.Ingest/Mapper/TableMapper.cs
```diff
@@ -1008,4 +1010,69 @@ public class TableMapper : TypeMapper<TableMetadata> {
...
+        results = MergeIndividualItemDropTable(results, parser.ParseIndividualItemGearBox());
+        results = MergeIndividualItemDropTable(results, parser.ParseIndividualItemDropEventNpc());
+        results = MergeIndividualItemDropTable(results, parser.ParseIndividualItemDropNewGacha());
+        results = MergeIndividualItemDropTable(results, parser.ParseIndividualItemDropQuestMob());
+        results = MergeIndividualItemDropTable(results, parser.ParseIndividualItemDropQuestObj());
+
+        return new IndividualItemDropTable(results);
+    }
+    
+    private Dictionary<int, Dictionary<byte, IList<IndividualItemDropTable.Entry>>> MergeIndividualItemDropTable(Dictionary<int, Dictionary<byte, IList<IndividualItemDropTable.Entry>>> results, IEnumerable<(int Id, IDictionary<byte,List<IndividualItemDrop>>)> parser) {
```
kOchirasu, line 1030: nit: `IDictionary<byte,List<IndividualItemDrop>>` => `IDictionary<byte, List<IndividualItemDrop>>`

Maple2.File.Ingest/Mapper/TableMapper.cs
kOchirasu: yea seems fine, not sure what I was thinking

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -72,6 +72,13 @@ public class ItemUseHandler : PacketHandler<GameSession> {
             case ItemFunction.VIPCoupon:
                 HandlePremiumClubCoupon(session, item);
                 break;
+            case ItemFunction.SelectItemBox:
+                HandleSelectItemBox(session, packet, item);
+                break;
+            case ItemFunction.OpenItemBox:
```
kOchirasu, line 78: `OpenItemBoxLullu` and `OpenItemBoxLulluSimple` can probably be included as well.

Maple2.Server.Game/PacketHandlers/ItemBoxHandler.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Maple2.Server.Game.PacketHandlers;
+
+public class ItemBoxHandler : PacketHandler<GameSession> {
+    public override RecvOp OpCode => RecvOp.RequestItemBox;
+
+    public override void Handle(GameSession session, IByteReader packet) {
+        int itemId = packet.ReadInt();
+        short unk = packet.ReadShort();
+        int count = packet.ReadInt();
+        short unk2 = packet.ReadShort(); // is 1 on select item box, 0 on open item box
```
kOchirasu, line 20: This is `UnicodeString` probably similar to the string in `RequestItemUse`

Maple2.Server.Game/Manager/ItemBoxManager.cs
```diff
@@ -0,0 +1,259 @@
...
+        }
+
+        // SelectItemBox disregards DropGroup
+        IList<IndividualItemDropTable.Entry> drops = session.ItemBox.FilterDrops(dropGroupTable.Values.SelectMany(x => x).ToList());
+
+        if (drops.Count == 0) {
+            return ItemBoxError.s_err_cannot_open_multi_itembox_inventory_fail;
+        }
+
+        IndividualItemDropTable.Entry selectedEntry = drops[index];
```
kOchirasu, line 57: This seems unsafe, probably should use `ElementAtOrDefault` to handle out of bounds. Can just do this single null check instead of the above `drops.Count == 0` check.

Maple2.Server.Game/Manager/ItemBoxManager.cs
```diff
@@ -0,0 +1,259 @@
...
+        for (int startCount = 0; startCount < count; startCount++) {
+            if (!session.Item.Inventory.ConsumeItemComponents(new[] {
```
kOchirasu, line 59: Here and below, since the array of `ItemComponent` is unchanged between iterations. I think it's better to define it once before the loop and reuse.

Maple2.Server.Game/Manager/ItemBoxManager.cs
```diff
@@ -0,0 +1,259 @@
...
+            for (int startCount = 0; startCount < count; startCount++) {
+                if (!session.Item.Inventory.ConsumeItemComponents(new[] {
+                        new ItemComponent(item.Id, item.Rarity, itemRequiredAmount, ItemTag.None),
+                    })) {
+                    return ItemBoxError.s_err_cannot_open_multi_itembox_inventory_fail;
+                }
+
+                if (itemId > 0 && session.ItemMetadata.TryGet(itemId, out ItemMetadata? itemMetadata)) {
+                    if (!session.Item.Inventory.Add(new Item(itemMetadata, item.Metadata.Option?.ConstantId ?? 1), true, true)) {
+                        error = ItemBoxError.s_err_cannot_open_multi_itembox_inventory;
```
kOchirasu, line 112: just to double check, we intentionally continue when we fail here?

Maple2.Server.Game/Manager/ItemBoxManager.cs
```diff
@@ -0,0 +1,259 @@
...
+            return ItemBoxError.s_err_cannot_open_multi_itembox_inventory_fail;
+        }
+
+        // check for keys
+        if (session.Item.Inventory.Find(keyItemId).Sum(requiredItem => requiredItem.Amount) < keyAmountRequired) {
+            return ItemBoxError.s_err_cannot_open_multi_itembox_inventory_fail;
+        }
+
+        // Get dropbox
+        return !session.TableMetadata.IndividualItemDropTable.Entries.TryGetValue(boxId, out Dictionary<byte, IList<IndividualItemDropTable.Entry>>? dropGroupTable)
```
kOchirasu, line 181: this is a bit dank, just use an if statement

Maple2.Server.Game/Manager/Items/InventoryManager.cs
kOchirasu: There's no precedence to use this dependency, just iterate to add:

```cs
for (Item item in session.Item.Inventory.Find(ingredient.ItemId, ingredient.Rarity)) {
    value.Add(item);
}
```

Maple2.Server.Game/Manager/Items/InventoryManager.cs
```diff
@@ -172,6 +173,9 @@ public class InventoryManager {
 
             IList<(Item, int Added)> result = items.Add(add, true);
             if (result.Count == 0) {
+                if (mailIfFull) {
+                    MailItem(add);
```
kOchirasu, line 177: I feel like returning true makes more sense after mailing, do we need to return false to handle something?

Maple2.Server.Game/Manager/ItemBoxManager.cs
Zintixx: Yeah the entire box opening process needs to finish before failing/stopping

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
Zintixx: I was gonna include this in a separate PR, along with ducky capsules and style crates

Maple2.Server.Game/Manager/Items/InventoryManager.cs
Zintixx: yes. I need a way to differentiate between if it failed to add to the inventory and mailed. if i had it set to false, it would return the same if it just mailed. Take for example line 247 in ItemBoxManager. I need to know if it failed so I can define the error to stop the box opening process later in the loop

kOchirasu: hmm, that's difficult because I feel like there needs to be a better way to differentiate between failure and mailed items. Maybe we should just leave `Add` as-is and expose `MailItem`. So usage would be something like:

```cs
if (!inventory.Add(item)) {
    inventory.MailItem(item);
}
```

Maple2.Server.Game/Manager/ItemBoxManager.cs
Zintixx: what do you mean? some are intentionally returning in order to prevent the function from continuing

kOchirasu: the variable is never used until after L68 so define it later.

Maple2.Server.Game/Manager/Items/InventoryManager.cs
Zintixx: wouldn't this create the item twice?

kOchirasu: Yea good point, perhaps we can add a check for `OpenSlots <= 0` and then return early without creating item in DB. If an item cannot stack, and there's no slots then we know the add will fail so don't need to create in DB yet.

Maple2.Server.Game/Manager/Items/InventoryManager.cs
```diff
@@ -154,20 +154,20 @@ public class InventoryManager {
...
 
+            if (!CanAdd(add)) {
+                return false;
+            }
+
             // If we are adding an item without a Uid, it may need to be created in db.
             if (add.Uid == 0) {
                 // Slot MUST be -1 so we don't add directly to a slot.
                 add.Slot = -1;
-                int remainStack = items.GetStackResult(add);
```
kOchirasu, line 161: you shouldn't remove this. If the item can fully stack into existing slots we don't need to create in DB.

Maple2.Server.Game/Manager/Items/InventoryManager.cs
```diff
@@ -154,20 +154,20 @@ public class InventoryManager {
                 return false;
             }
 
+            if (!CanAdd(add)) {
```
kOchirasu, line 157: I would prefer just checking `OpenSlots` in conjunction with the check below because `CanAdd` is checking multiple things which are already being checked. (e.g. ItemCollection above, and GetStackResult below)

Maple2.Server.Game/Manager/Items/InventoryManager.cs
```diff
@@ -333,6 +339,41 @@ public class InventoryManager {
         return true;
     }
 
+    public bool MailItem(Item item) {
```
kOchirasu, line 342: I would move this to `ItemManager` since it's not really inventory related anymore. Also you don't need the lock since it doesn't mutate inventory.

Maple2.Server.Game/Manager/Items/InventoryManager.cs
```diff
@@ -333,6 +339,41 @@ public class InventoryManager {
...
+            using GameStorage.Request db = session.GameStorage.Context();
+            var mail = new Mail {
+                Type = MailType.System,
+                ReceiverId = session.CharacterId,
+                Content = "50000000", // id from string/en/systemmailcontentna.xml
+            };
+
+            mail = db.CreateMail(mail);
+            if (mail == null) {
+                throw new InvalidOperationException($"Failed to create mail for character id: {session.CharacterId}");
```
kOchirasu, line 353: can just return `false` here, we are not in a bad state at this point.

Maple2.Server.Game/Manager/Items/InventoryManager.cs
```diff
@@ -159,15 +159,16 @@ public class InventoryManager {
...
                 int remainStack = items.GetStackResult(add);
-                if (remainStack > 0) {
-                    using GameStorage.Request db = session.GameStorage.Context();
-                    Item? newAdd = db.CreateItem(session.CharacterId, add);
-                    if (newAdd == null) {
-                        return false;
-                    }
-
-                    add = newAdd;
+                if (remainStack == 0 && items.OpenSlots == 0) {
```
kOchirasu, line 162: This is still wrong. If `remainStack` is 0, we CAN add but do not stack. the full logic should be something like:

```cs
int remainStack = items.GetStackResult(add);
if (remainStack > 0) {
    if (items.OpenSlots <= 0) {
        return false;
    }

    using GameStorage.Request db = session.GameStorage.Context();
    Item? newAdd = db.CreateItem(session.CharacterId, add);
    if (newAdd == null) {
        return false;
    }

    add = newAdd;
}
```

PR #30, opened 2023-04-03
Add starting equipment to player inventory in tutorial

julianvu:
This PR adds the `TutorialItemHandler` that spawns the starting equipment belonging to the class of the player, allowing the player to progress beyond the tutorial.

### DIFF
--- a/Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
+++ b/Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
@@ -34,7 +34,9 @@ public class TutorialItemHandler : PacketHandler<GameSession> {
 
             tutorialItemsInInventory += session.Item.Equips.Gear.Count(x => x.Value.Id == tutorialItemGroup.Key);
 
-            if (tutorialItemsInInventory >= tutorialItemGroup.Count()) {
+            int tutorialItemsToSpawn = tutorialItemGroup.Select(item => item.Count).Sum();
+
+            if (tutorialItemsInInventory >= tutorialItemsToSpawn) {
                 continue;
             }
 
@@ -42,8 +44,9 @@ public class TutorialItemHandler : PacketHandler<GameSession> {
                 continue;
             }
 
-            foreach (var tutorialItem in tutorialItemGroup) {
-                session.Item.Inventory.Add(new Item(itemMetadata, tutorialItem.Rarity));
+            while (tutorialItemsInInventory < tutorialItemsToSpawn) {
+                session.Item.Inventory.Add(new Item(itemMetadata));
+                tutorialItemsInInventory++;
             }
         }
     }
--- a/Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
+++ b/Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
@@ -32,7 +32,7 @@ public class TutorialItemHandler : PacketHandler<GameSession> {
 
             if (!session.ItemMetadata.TryGet(tutorialItem.Id, out var itemMetadata)) continue;
 
-            Item item = new(itemMetadata, tutorialItem.Rarity);
+            var item = new Item(itemMetadata, tutorialItem.Rarity);
             session.Item.Inventory.Add(item, true);
         }
     }
--- a/Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
+++ b/Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
@@ -13,27 +13,38 @@ public class TutorialItemHandler : PacketHandler<GameSession> {
     public override RecvOp OpCode => RecvOp.RequestTutorialItem;
 
     public override void Handle(GameSession session, IByteReader packet) {
+        if (session.Player.Value.Character.Level > 1) {
+            return;
+        }
+
         var jobTable = session.TableMetadata.JobTable;
 
         var jobCode = session.Player.Value.Character.Job.Code();
 
-        if (!jobTable.Entries.TryGetValue(jobCode, out var jobMetadata)) return;
-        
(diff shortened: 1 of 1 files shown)


Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+using Maple2.Server.Core.PacketHandlers;
+using Maple2.Server.Game.Session;
+
+namespace Maple2.Server.Game.PacketHandlers;
+
+public class TutorialItemHandler : PacketHandler<GameSession>
+{
+    public override RecvOp OpCode => RecvOp.RequestTutorialItem;
+
+    public override void Handle(GameSession session, IByteReader packet)
```
kOchirasu, line 17: I would do some additional checks here to prevent people from exploiting this function.

Level check: must be level 1
Map check: must be in starting map
- this info is available from `JobTable.Entry.Tutorial.StartField`

Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                continue;
+            }
+
+            if (!session.ItemMetadata.TryGet(tutorialItem.Id, out ItemMetadata? itemMetadata))
+            {
+                continue;
+            }
+
+            Item item = new(itemMetadata, tutorialItem.Rarity);
```
kOchirasu, line 46: for consistency with rest of the project:
```cs 
var item = new Item(...);
```

Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+        JobTable jobTable = session.TableMetadata.JobTable;
+
+        JobCode jobCode = session.Player.Value.Character.Job.Code();
+
+        if (!jobTable.Entries.TryGetValue(jobCode, out JobTable.Entry? jobMetadata))
+        {
+            return;
+        }
+
+        foreach (JobTable.Item tutorialItem in jobMetadata.Tutorial.StartItem)
```
kOchirasu, line 28: I don't like that for classes like thief we are still running this loop twice even though the first iteration already rewards all start items. How about doing a `GroupBy` on `StartItem` by `Id` so that we can loop over just unique ids. For each group you can sum up the count to get the total items needed.

Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+using Maple2.Model.Game;
+using Maple2.Model.Metadata;
+using Maple2.PacketLib.Tools;
+using Maple2.Server.Core.Constants;
+using Maple2.Server.Core.PacketHandlers;
+using Maple2.Server.Game.Session;
+
+namespace Maple2.Server.Game.PacketHandlers;
+
+public class TutorialItemHandler : PacketHandler<GameSession>
```
kOchirasu, line 13: Formatting: we use K&R brace style.

Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
julianvu: Good call 👍

Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+
+public class TutorialItemHandler : PacketHandler<GameSession> {
+    public override RecvOp OpCode => RecvOp.RequestTutorialItem;
+
+    public override void Handle(GameSession session, IByteReader packet) {
+        var jobTable = session.TableMetadata.JobTable;
+
+        var jobCode = session.Player.Value.Character.Job.Code();
+
+        if (!jobTable.Entries.TryGetValue(jobCode, out var jobMetadata)) return;
```
kOchirasu, line 20: Here and below, for consistency: Always use braces, even for single-line statements. (I think there are very few exceptions)

Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+        
+        if (session.Field != null && (session.Player.Value.Character.Level > 1 || jobMetadata.Tutorial.StartField != session.Field.MapId)) return;
```
kOchirasu, line 22: nit: I would check `Level` up top to return as early as possible. the `StartField` check can remain here since we need `jobMetadata` first.

Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
julianvu: Could you check what I did for this? I'm not sure it's what you're expecting 😅

Maple2.Server.Game/PacketHandlers/TutorialItemHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            if (tutorialItemsInInventory >= tutorialItemGroup.Count()) {
+                continue;
+            }
+
+            if (!session.ItemMetadata.TryGet(tutorialItemGroup.Key, out var itemMetadata)) {
+                continue;
+            }
+
+            foreach (var tutorialItem in tutorialItemGroup) {
```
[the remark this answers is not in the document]
kOchirasu: Something like this may be cleaner:
```cs
int totalItems = tutorialItemGroup.Select(item => item.Count).Sum();
while (tutorialItemsInInventory < totalItems) {
    session.Item.Inventory.Add(new Item(itemMetadata, tutorialItem.Rarity));
    tutorialItemsInInventory++;
}
```

PR #32, opened 2023-04-04
Job, level and currentMap functions

gwagwaflamingo:
WIP queststate functions commented out for the future.

### DIFF
--- a/Maple2.Server.Game/Scripting/Npc/NpcScriptContext.cs
+++ b/Maple2.Server.Game/Scripting/Npc/NpcScriptContext.cs
@@ -278,6 +278,6 @@ public class NpcScriptContext {
     } */
 
     public int CurrentMap() {
-        return Session.Field.MapId;
+        return Session.Field?.MapId ?? 0;
     }
 }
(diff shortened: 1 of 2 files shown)


Maple2.Server.Game/Scripting/Npc/NpcScriptContext.cs
```diff
@@ -255,4 +255,29 @@ public class NpcScriptContext {
...
+    public bool MultiQuestState(List<int> questIds, int questState) {
+        foreach (int questId in questIds) {
+            if (Session.Quest.Find(questId).State == questState) {
+                return true;
+            }
+        }
+    }
+    } */
+
+    public bool CurrentMap(int mapId) {
```
kOchirasu, line 280: Seems more flexible to not pass `mapId` here, just return `Session.Field.MapId` and you can check against `mapId` in the script.

Maple2.Server.Game/Scripting/Npc/NpcScriptContext.cs
```diff
@@ -255,4 +255,29 @@ public class NpcScriptContext {
...
+        return (int)Session.Player.Value.Character.Job;
+    }
+
+    public int Level() {
+        return Session.Player.Value.Character.Level;
+    }
+
+    /* TODO: Quests aren't implemented yet :( Finish once they are, I guess.
+    public bool QuestState(int questId, int questState) {
+        return Session.Quest.Find(questId).State == questState;
```
kOchirasu, line 269: braces got messed up for this function

gwagwaflamingo: Thanks, fixed it 👍

PR #35, opened 2023-04-06
Item Color

Zintixx:
This includes a new CreateItem function that will need to replace the entirety of all Item constructors with this function. Currently it's just implemented on the ItemCommand for proof of concept.

EDIT: good to merge!

### DIFF
--- a/Maple2.Model/Metadata/Table/ColorPaletteTable.cs
+++ b/Maple2.Model/Metadata/Table/ColorPaletteTable.cs
@@ -3,7 +3,7 @@ using Maple2.Model.Common;
 
 namespace Maple2.Model.Metadata;
 
-public record ColorPaletteTable(Dictionary<int, Dictionary<int, ColorPaletteTable.Entry>> Entries) : Table {
+public record ColorPaletteTable(IReadOnlyDictionary<int, Dictionary<int, ColorPaletteTable.Entry>> Entries) : Table {
 
     public record Entry(Color Primary,
                         Color Secondary,
--- a/Maple2.Model/Metadata/Table/ColorPaletteTable.cs
+++ b/Maple2.Model/Metadata/Table/ColorPaletteTable.cs
@@ -1,0 +1,13 @@
+﻿using System.Collections.Generic;
+using Maple2.Model.Common;
+
+namespace Maple2.Model.Metadata;
+
+public record ColorPaletteTable(Dictionary<int, Dictionary<int, ColorPaletteTable.Entry>> Entries) : Table {
+
+    public record Entry(Color Primary,
+                        Color Secondary,
+                        Color Tertiary,
+                        int AchieveId,
+                        int AchieveGrade);
+}
--- a/Maple2.Model/Metadata/Table/ColorPaletteTable.cs
+++ b/Maple2.Model/Metadata/Table/ColorPaletteTable.cs
@@ -3,7 +3,7 @@ using Maple2.Model.Common;
 
 namespace Maple2.Model.Metadata;
 
-public record ColorPaletteTable(IReadOnlyDictionary<int, Dictionary<int, ColorPaletteTable.Entry>> Entries) : Table {
+public record ColorPaletteTable(IReadOnlyDictionary<int, IReadOnlyDictionary<int, ColorPaletteTable.Entry>> Entries) : Table {
 
     public record Entry(Color Primary,
                         Color Secondary,
--- a/Maple2.Server.Game/Manager/Items/ItemManager.cs
+++ b/Maple2.Server.Game/Manager/Items/ItemManager.cs
@@ -50,7 +50,7 @@ public class ItemManager {
     }
 
     public Item CreateItem(ItemMetadata itemMetadata, int rarity = 1, int amount = 1) {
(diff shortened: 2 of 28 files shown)


Maple2.Model/Metadata/Table/ColorPaletteTable.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
+using Maple2.Model.Common;
+
+namespace Maple2.Model.Metadata;
+
+public record ColorPaletteTable(Dictionary<int, Dictionary<int, ColorPaletteTable.Entry>> Entries) : Table {
```
kOchirasu, line 6: IReadOnlyDictionary

Maple2.Server.Game/Manager/Items/ItemManager.cs
```diff
@@ -42,6 +49,45 @@ public class ItemManager {
         return item ?? Equips.Outfit.Values.FirstOrDefault(outfit => outfit.Uid == uid);
     }
 
+    public Item CreateItem(ItemMetadata itemMetadata, int rarity = 1, int amount = 1, bool initialize = true) {
+        var item = new Item(itemMetadata, rarity, amount, initialize);
+        item.Stats = itemStatsCalc.GetStats(item);
+        item.Socket = itemStatsCalc.GetSockets(item);
+
```
kOchirasu, line 56: nit: extra line break

Maple2.Server.Game/Manager/Items/ItemManager.cs
```diff
@@ -42,6 +49,45 @@ public class ItemManager {
         return item ?? Equips.Outfit.Values.FirstOrDefault(outfit => outfit.Uid == uid);
     }
 
+    public Item CreateItem(ItemMetadata itemMetadata, int rarity = 1, int amount = 1, bool initialize = true) {
+        var item = new Item(itemMetadata, rarity, amount, initialize);
+        item.Stats = itemStatsCalc.GetStats(item);
+        item.Socket = itemStatsCalc.GetSockets(item);
+        
+        
+        item.Appearance = itemMetadata.SlotNames.FirstOrDefault(EquipSlot.Unknown) switch {
```
kOchirasu, line 58: Looks like `item.Appearance` is mutable, so you can just do:
```cs
item.Appearance.Color = GetColor(item);
```

The appropriate Appearance type is already initialized in the constructor.

Maple2.Server.Game/Manager/Items/ItemManager.cs
```diff
@@ -42,6 +49,45 @@ public class ItemManager {
         return item ?? Equips.Outfit.Values.FirstOrDefault(outfit => outfit.Uid == uid);
     }
 
+    public Item CreateItem(ItemMetadata itemMetadata, int rarity = 1, int amount = 1, bool initialize = true) {
```
kOchirasu, line 52: would not allow passing `initialize` here. Pretty sure it should always be `true`. `false` is really only used for loading from DB, in which case we don't need to call this function since stats/color have already been initialized.

Maple2.Server.Game/Manager/Items/ItemManager.cs
```diff
@@ -42,6 +49,45 @@ public class ItemManager {
...
+            EquipSlot.HR => new HairAppearance(GetColor(item)),
+            EquipSlot.FD => new DecalAppearance(GetColor(item)),
+            EquipSlot.CP => new CapAppearance(GetColor(item)),
+            _ => new ItemAppearance(GetColor(item)),
+        };
+        
+        return item;
+    }
+    
+    private EquipColor GetColor(Item item) {
```
kOchirasu, line 68: maybe pass `item.Metadata.Customize` rather than `item`

Maple2.Model/Metadata/Table/ColorPaletteTable.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
+using Maple2.Model.Common;
+
+namespace Maple2.Model.Metadata;
+
+public record ColorPaletteTable(IReadOnlyDictionary<int, Dictionary<int, ColorPaletteTable.Entry>> Entries) : Table {
```
kOchirasu, line 6: Other `Dictionary` should also be `IReadOnlyDictionary`

Zintixx: not sure if i would be able to do that with how i'm ingesting it.

kOchirasu: you should change the way you're ingesting it to make this possible (build a separate dictionary, then insert it once it's done being built). It's important that all metadata is fully immutable because a single instance is shared across the whole server.

Maple2.Server.Game/PacketHandlers/AttendanceHandler.cs
```diff
@@ -94,6 +94,11 @@ private enum Command : byte {
             return;
```
kOchirasu, line 94: `ItemMetadata` lookup above can be omitted

PR #36, opened 2023-06-02
Item Repack

Zintixx:
I changed "CanRepack" in `ItemEnchant` to "Tradeable." Open for discussion on a better name for this without sounding misleading. This bool is true if
~~`Transfer.RepackageCount > 0 || Transfer.RemainTrades > 0 || Transfer.Flag.HasFlag(TransferFlag.Trade)`~~
`Transfer.Flag.HasFlag(TransferFlag.Trade) || Transfer.Flag.HasFlag(TransferFlag.LimitTrade)`

### DIFF
--- a/Maple2.Server.Game/PacketHandlers/ItemRepackHandler.cs
+++ b/Maple2.Server.Game/PacketHandlers/ItemRepackHandler.cs
@@ -39,7 +39,7 @@ public class ItemRepackHandler : PacketHandler<GameSession> {
             }
 
             Item? targetItem = session.Item.Inventory.Get(targetItemUid);
-            if (targetItem == null || targetItem.Binding != null) {
+            if (targetItem == null || targetItem.Transfer == null || targetItem.Binding != null) {
                 session.Send(ItemRepackPacket.Error(ItemRepackError.s_item_repacking_scroll_error_invalid_target));
                 return;
             }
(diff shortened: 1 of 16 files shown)


kOchirasu:
Do we need a db migration so old items don't break with this rename? (Or if the field wasn't really used before it might be ok)

Zintixx:
I tried doing it but considering it's in a json blob, it didn't generate a db migration.

Maple2.Server.Game/PacketHandlers/ItemRepackHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        lock (session.Item) {
+            Item? scroll = session.Item.Inventory.Get(scrollUid);
+            if (scroll == null) {
+                session.Send(ItemRepackPacket.Error(ItemRepackError.s_item_repacking_scroll_error_invalid_scroll));
+                return;
+            }
+
+            Item? targetItem = session.Item.Inventory.Get(targetItemUid);
+            if (targetItem == null || targetItem.Binding != null) {
```
kOchirasu, line 42: seems like we should just check `targetItem.Transfer != null` here as well to simplify logic below.

Zintixx: Do you mean it should be 
`if (targetItem == null || targetItem.Transfer != null || targetItem.Binding != null)` ?
I think I'd have to put this in another if statement or have the check on Binding in another statement otherwise it would allow it to pass through due to Binding

kOchirasu: i think it should be:
```cs
if (targetItem == null || targetItem.Transfer == null || targetItem.Binding != null)
```

Maple2.Server.Game/PacketHandlers/ItemRepackHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+                session.Send(ItemRepackPacket.Error(ItemRepackError.s_item_repacking_scroll_error_invalid_target));
+                return;
+            }
+
+            if (targetItem.Metadata.Property.RepackConsumeCount > scroll.Amount) {
+                session.Send(ItemRepackPacket.Error(ItemRepackError.s_item_repacking_scroll_error_server_fail_consume_scroll));
+                return;
+            }
+
+            if (targetItem.Transfer.Flag.HasFlag(TransferFlag.LimitTrade) == true && targetItem.Transfer.RemainTrades > 0) {
```
kOchirasu, line 52: you don't need `== true` because it's not nullable anymore

PR #37, opened 2023-06-04
Gacha Capsules & Style Crates/Lullu Boxes

Zintixx:


Maple2.Server.Game/Manager/ItemBoxManager.cs
```diff
@@ -29,11 +31,120 @@ public class ItemBoxManager {
...
+        var totalItems = new List<Item>();
+        for (int startCount = 0; startCount < count; startCount++) {
+            if (!session.Item.Inventory.Consume(new[] {keyIngredient})) {
+                return ItemBoxError.s_err_cannot_open_multi_itembox_inventory;
+            }
+
+            if (autoPay) {
+                int.TryParse(parameters["boxPrice"], out int mesoCost);
+                if (!session.Item.Inventory.Consume(new[] {boxIngredient})
+                    && mesoCost > 0 && session.Currency.CanAddMeso(-mesoCost) == -mesoCost) {
```
kOchirasu, line 65: mesos should be checked before consume is called because consume will mutate the inventory.

Might be simpler to change it to something like:
```cs
if (autoPay && mesoCost > 0 && session.Currency.CanAddMeso(-mesoCost) != -mesoCost) {
    return ItemBoxError.s_err_cannot_open_multi_itembox_inventory;
}

if (!session.Item.Inventory.Consume(new[] {boxIngredient})) {
    return ItemBoxError.s_err_cannot_open_multi_itembox_inventory;
}
```

Maple2.Server.Game/Packets/ItemScriptPacket.cs
```diff
@@ -0,0 +1,51 @@
...
+    
+    public static ByteWriter LulluBox(ICollection<Item> items) {
+        var pWriter = Packet.Of(SendOp.ItemScript);
+        pWriter.Write<Command>(Command.LulluBox);
+        pWriter.WriteInt(items.Count);
+        foreach (Item item in items) {
+            pWriter.WriteInt(item.Id);
+            pWriter.WriteInt(item.Amount);
+            pWriter.WriteInt(item.Rarity);
+            pWriter.WriteByte();
```
kOchirasu, line 24: im curious what this value is, can you try setting it to 1 and see if it displays any different? (might be a bool to display new or log in chat or something)

Zintixx: ahhh this is just for the little sparkling animation around the item

set at false:
false

true

Referencing this video it looks like it's always set to true regardless of the item
https://www.youtube.com/watch?v=laxvp4M8i3Y

Maple2.Server.Game/Manager/ItemBoxManager.cs
Zintixx: The way it works is it's auto pay, it'll consume the boxIngredient first until needing to pay mesos for it.
So if I set it to auto pay for 10 boxes and I only have 6 boxes, it'll consume the 6 boxes first then pay for the remaining 4 in mesos. It needs to consume the ingredients first before checking mesos and paying

Zintixx: I realized it's not actually paying the mesos.
I pushed a change, let me know what you think. I'm doing Math.Min because sometimes the value is -1 (aka no cost)

Maple2.Server.Game/Packets/ItemScriptPacket.cs
kOchirasu: Can you add a comment for this

Maple2.Server.Game/Manager/ItemBoxManager.cs
kOchirasu: Looks good, but I think it's supposed to be `!=` for mesoCost

PR #39, opened 2023-07-30
WIP: Buffs part 1

Zintixx:
- Added BuffManager 
- Added prelim support for item and map buffs.

### DIFF
--- a/Maple2.File.Ingest/Mapper/AdditionalEffectMapper.cs
+++ b/Maple2.File.Ingest/Mapper/AdditionalEffectMapper.cs
@@ -46,6 +46,7 @@ public class AdditionalEffectMapper : TypeMapper<AdditionalEffectMetadata> {
                     Consume: new AdditionalEffectMetadataConsume(
                         HpRate: data.ConsumeProperty.hpRate,
                         SpRate: data.ConsumeProperty.spRate),
+                    Reflect: Convert(data.ReflectProperty),
                     Update: Convert(data),
                     Status: Convert(data.StatusProperty, data.OffensiveProperty, data.DefensiveProperty),
                     Recovery: Convert(data.RecoveryProperty),
@@ -82,6 +83,26 @@ public class AdditionalEffectMapper : TypeMapper<AdditionalEffectMetadata> {
             ImmuneCategories: Array.ConvertAll(data.ImmuneEffectProperty.immuneBuffCategories, category => (BuffCategory) category),
             ResetCooldown: data.ResetSkillCoolDownTimeProperty.skillCodes,
             Duration: modifyDuration);
+    }
+
+    private static AdditionalEffectMetadataReflect Convert(ReflectProperty reflect) {
+        var values = new Dictionary<BasicAttribute, long>();
+        var rates = new Dictionary<BasicAttribute, float>();
+        
+        values.AddIfNotDefault(BasicAttribute.PhysicalAtk, reflect.physicalReflectionValue);
+        values.AddIfNotDefault(BasicAttribute.MagicalAtk, reflect.magicalReflectionValue);
+        
+        rates.AddIfNotDefault(BasicAttribute.PhysicalAtk, reflect.physicalReflectionRate);
+        rates.AddIfNotDefault(BasicAttribute.MagicalAtk, reflect.magicalReflectionRate);
+        return new AdditionalEffectMetadataReflect(
+            reflect.reflectionRate,
+            reflect.reflectionAdditionalEffectId,
+            reflect.reflectionAdditionalEffectLevel,
+            reflect.reflectionCount,
+            reflect.physicalReflectionRateLimit,
+            reflect.magicalReflectionRateLimit,
+            values,
+            rates);
     }
 
     private static AdditionalEffectMetadataStatus Convert(StatusProperty status, OffensiveProperty offensive, DefensiveProperty defensive) {
--- a/Maple2.File.Ingest/Mapper/AdditionalEffectMapper.cs
+++ b/Maple2.File.Ingest/Mapper/AdditionalEffectMapper.cs
@@ -34,7 +34,7 @@ public class AdditionalEffectMapper : TypeMapper<AdditionalEffectMetadata> {
                         DelayTick: data.BasicProperty.delayTick,
                         MaxCount: data.BasicProperty.maxBuffCount,
                         KeepCondition: (BuffKeepCondition) data.BasicProperty.keepCondition,
-                        ResetCondition: (BuffKeepCondition) data.BasicProperty.resetCondition,
+                        ResetCondition: (BuffResetCondition) data.BasicProperty.resetCondition,
                         DotCondition: (BuffDotCondition) data.BasicProperty.dotCondition),
                     Consume: new AdditionalEffectMetadataConsume(
                         HpRate: data.ConsumeProperty.hpRate,
--- a/Maple2.File.Ingest/Mapper/AdditionalEffectMapper.cs
+++ b/Maple2.File.Ingest/Mapper/AdditionalEffectMapper.cs
@@ -39,6 +39,7 @@ public class AdditionalEffectMapper : TypeMapper<AdditionalEffectMetadata> {
                         RemoveOnPvpZone: data.BasicProperty.clearEffectFromPVPZone,
                         KeepOnEnterPvpZone: data.BasicProperty.doNotClearEffectFromEnterPVPZone,
                         CasterIndividualBuff: data.BasicProperty.casterIndividualEffect,
+                        Exp: data.ExpProperty.value,
                         KeepCondition: (BuffKeepCondition) data.BasicProperty.keepCondition,
                         ResetCondition: (BuffResetCondition) data.BasicProperty.resetCondition,
                         DotCondition: (BuffDotCondition) data.BasicProperty.dotCondition),
@@ -51,8 +52,6 @@ public class AdditionalEffectMapper : TypeMapper<AdditionalEffectMetadata> {
                     Dot: new AdditionalEffectMetadataDot(
                         Damage: Convert(data.DotDamageProperty),
                         Buff: Convert(data.DotBuffProperty)),
-                    Exp: new AdditionalEffectMetadataExp(
-                        Value: data.ExpProperty.value),
                     Shield: Convert(data.ShieldProperty),
                     InvokeEffect: Convert(data.InvokeEffectProperty),
                     Skills: data.conditionSkill.Concat(data.splashSkill).Select(skill => skill.Convert()).ToArray());
--- a/Maple2.File.Ingest/Mapper/AdditionalEffectMapper.cs
+++ b/Maple2.File.Ingest/Mapper/AdditionalEffectMapper.cs
@@ -33,6 +33,12 @@ public class AdditionalEffectMapper : TypeMapper<AdditionalEffectMetadata> {
                         IntervalTick: data.BasicProperty.intervalTick,
                         DelayTick: data.BasicProperty.delayTick,
                         MaxCount: data.BasicProperty.maxBuffCount,
+                        KeepOnDeath: data.BasicProperty.deadKeepEffect,
+                        RemoveOnLogout: data.BasicProperty.logoutClearEffect,
+                        RemoveOnLeaveField: data.BasicProperty.leaveFieldClearEffect, 
+                        RemoveOnPvpZone: data.BasicProperty.clearEffectFromPVPZone,
+                        KeepOnEnterPvpZone: data.BasicProperty.doNotClearEffectFromEnterPVPZone,
+                        CasterIndividualBuff: data.BasicProperty.casterIndividualEffect,
                         KeepCondition: (BuffKeepCondition) data.BasicProperty.keepCondition,
                         ResetCondition: (BuffResetCondition) data.BasicProperty.resetCondition,
                         DotCondition: (BuffDotCondition) data.BasicProperty.dotCondition),
@@ -45,6 +51,8 @@ public class AdditionalEffectMapper : TypeMapper<AdditionalEffectMetadata> {
                     Dot: new AdditionalEffectMetadataDot(
                         Damage: Convert(data.DotDamageProperty),
                         Buff: Convert(data.DotBuffProperty)),
+                    Exp: new AdditionalEffectMetadataExp(
+                        Value: data.ExpProperty.value),
                     Shield: Convert(data.ShieldProperty),
                     InvokeEffect: Convert(data.InvokeEffectProperty),
                     Skills: data.conditionSkill.Concat(data.splashSkill).Select(skill => skill.Convert()).ToArray());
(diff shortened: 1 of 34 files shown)


Maple2.Server.Game/Manager/Config/BuffManager.cs
```diff
@@ -0,0 +1,147 @@
...
+    private int idCounter;
+
+    /// <summary>
+    /// Generates an ObjectId unique to this specific actor instance.
+    /// </summary>
+    /// <returns>Returns a local ObjectId</returns>
+    private int NextLocalId() => Interlocked.Increment(ref idCounter);
+    #endregion
+    private readonly IActor actor;
+    public virtual ConcurrentDictionary<int, Buff> Buffs { get; } = new();
```
kOchirasu, line 23: this shouldn't be virtual

Maple2.Server.Game/Manager/Config/BuffManager.cs
```diff
@@ -0,0 +1,147 @@
...
+        // Load buffs that are not broadcasted to the field
+        if (actor is FieldPlayer player) {
+            player.Session.Config.Skill.UpdatePassiveBuffs(false);
+            foreach (Item item in player.Session.Item.Equips.Gear.Values) {
+                foreach (ItemMetadataAdditionalEffect buff in item.Metadata.AdditionalEffects) {
+                    AddBuff(actor, actor, buff.Id, buff.Level, false);
+                }
+
+                // Check for buffs on gemstones
+                if (item.Socket?.Sockets is not null && item.Socket.Sockets.Length > 0)
```
kOchirasu, line 40: this if statement is missing braces

Maple2.Server.Game/Model/Field/Actor/Actor.cs
```diff
@@ -17,16 +20,7 @@ namespace Maple2.Server.Game.Model;
...
-    #region ObjectId
-    private int idCounter;
-
-    /// <summary>
-    /// Generates an ObjectId unique to this specific actor instance.
-    /// </summary>
-    /// <returns>Returns a local ObjectId</returns>
-    protected int NextLocalId() => Interlocked.Increment(ref idCounter);
-    #endregion
+public abstract class Actor<T> : IActor<T>, IActor, IDisposable {
```
kOchirasu, line 23: do you need to add `IActor` here? `IActor<T>` is already here and `IActor<T>` is `IActor`

Maple2.File.Ingest/Mapper/AdditionalEffectMapper.cs
```diff
@@ -33,9 +33,9 @@ public class AdditionalEffectMapper : TypeMapper<AdditionalEffectMetadata> {
                         IntervalTick: data.BasicProperty.intervalTick,
                         DelayTick: data.BasicProperty.delayTick,
                         MaxCount: data.BasicProperty.maxBuffCount,
-                        KeepCondition: data.BasicProperty.keepCondition,
-                        ResetCondition: data.BasicProperty.resetCondition,
-                        DotCondition: data.BasicProperty.dotCondition),
+                        KeepCondition: (BuffKeepCondition) data.BasicProperty.keepCondition,
+                        ResetCondition: (BuffKeepCondition) data.BasicProperty.resetCondition,
```
mettaursp, line 37: Should this be a BuffResetCondition cast instead?

Maple2.Model/Metadata/AdditionalEffectMetadata.cs
```diff
@@ -27,9 +27,9 @@ public record AdditionalEffectMetadataProperty(
     int IntervalTick,
     int DelayTick,
     int MaxCount,
-    int KeepCondition,
-    int ResetCondition,
-    int DotCondition);
+    BuffKeepCondition KeepCondition,
+    BuffKeepCondition ResetCondition,
```
mettaursp, line 31: Should this be BuffResetCondition instead?

Maple2.Server.Game/Manager/Config/BuffManager.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+            return;
+        }
+
+        if (!owner.Field.SkillMetadata.TryGetEffect(id, level, out AdditionalEffectMetadata? additionalEffect)) {
+            logger.Error("Invalid buff: {SkillId},{Level}", id, level);
+            return;
+        }
+
+        // if (!SkillUtils.CheckCondition(additionalEffect.Condition, caster, owner, this)) {
```
mettaursp, line 80: For future reference, the buff should still be added, just disabled, if the condition isn't met. The disabled buffs usually stay dormant in the actor until the condition is met and then flip on. There should be an enabled/disabled flag in the packets so it will get sent too, but the state of the buff changes. For example: striker effect `11000282` has `level.beginCondition.owner.eventCondition="4"` which waits for when the player is attacked to enable, which switches the counter attack skill 11000281 with a `changeSkill` to skill `11000282`.

Maple2.Server.Game/Manager/Config/BuffManager.cs
mettaursp, line 1: When you can get around to it, it might be a good idea to start thinking about adding stubs for handling `eventCondition`.  There are a lot of enum values for it, and event sources are littered all throughout combat code. They all feed back into the active additional effects. The sooner you start thinking about it the easier it will be to integrate them.

The way I went about it in MapleServer2 was with `SkillTriggerHandler.FireEvents` calls placed in each of the systems that are referenced by the `eventCondition` enum values.

Maple2.Server.Game/Manager/Config/BuffManager.cs
```diff
@@ -0,0 +1,148 @@
...
+        //     Console.WriteLine($"Ignore buff {id}");
+        //     return;
+        // }
+
+        var buff = new Buff(owner.Field, additionalEffect, NextLocalId(), caster, owner);
+        if (!Buffs.TryAdd(id, buff)) {
```
mettaursp, line 86: Another note for the future: if `casterIndividualEffect="1"` then you should track who casted the effect & only check for buffs placed by the same caster. For example effect `10200112` - `Deep Cuts` can be applied by multiple berserkers simultaneously.

Maple2.Model/Metadata/AdditionalEffectMetadata.cs
```diff
@@ -73,6 +80,9 @@ public record AdditionalEffectMetadataRecovery(
     float EpRate,
     bool NotCrit);
 
+public record AdditionalEffectMetadataExp (
```
kOchirasu, line 83: separate type seems unnecessary, will there be other fields here in the future?

Maple2.Model/Metadata/BeginCondition.cs
```diff
@@ -7,10 +8,19 @@ public record BeginCondition(
     long Mesos,
     Gender Gender,
     JobCode[] JobCode,
+    float Probability,
+    bool OnlyShadowWorld,
+    bool OnlyFlyableMap,
+    Dictionary<BasicAttribute, long> Stat,
```
kOchirasu, line 14: `IReadOnlyDictionary`

Maple2.Model/Metadata/BeginCondition.cs
```diff
@@ -7,10 +8,19 @@ public record BeginCondition(
...
+    BeginConditionWeapon[]? Weapon,
     BeginConditionTarget? Target,
     BeginConditionTarget? Owner,
     BeginConditionTarget? Caster);
 
+public record BeginConditionWeapon(
```
kOchirasu, line 20: curious what these bytes even mean, should it be an enum maybe?

Maple2.Server.Game/Commands/BuffCommand.cs
```diff
@@ -0,0 +1,54 @@
...
+using Maple2.Server.Game.Packets;
+using Maple2.Server.Game.Session;
+
+namespace Maple2.Server.Game.Commands;
+
+public class BuffCommand : Command {
+    private const string NAME = "buff";
+    private const string DESCRIPTION = "Add buff to player.";
+
+    private const int MAX_RARITY = 6;
```
kOchirasu, line 16: remove these unused consts

Maple2.Server.Game/Manager/Config/BuffManager.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        //TODO: Check if buff is removable/should be removed
+        Buffs.Remove(id, out _);
+        actor.Field.Broadcast(BuffPacket.Remove(buff));
+        return true;
+
+    }
+
+    public void RemoveAll() {
```
kOchirasu, line 208: Do any packets need to be sent here?

Maple2.Server.Game/Model/Field/Buff.cs
```diff
@@ -93,42 +107,26 @@ public class Buff : IUpdatable, IByteSerializable {
...
-            Remove();
-            return;
-        }
-
-        // TODO: Check conditions less frequently
-        if (!UpdateEnabled()) {
-            return;
-        }
-
+    public void Proc(long tickCount) {
```
kOchirasu, line 116: why rename this from `Update`, seems clearer to keep these things with `IUpdatable`. A proc isn't guaranteed on every call of this function anyways (`tickCount < NextProcTick`)

Maple2.Server.Game/PacketHandlers/SkillHandler.cs
```diff
@@ -103,14 +105,24 @@ private enum SubCommand : byte {
 
         packet.ReadInt(); // ClientTick
         record.Unknown = packet.ReadBool(); // UnkBool
-        packet.ReadLong(); // UnkLong
+        long itemUid = packet.ReadLong(); // UnkLong
```
kOchirasu, line 108: nit: remove comment

Maple2.Server.Game/PacketHandlers/SkillHandler.cs
```diff
@@ -103,14 +105,24 @@ private enum SubCommand : byte {
...
         record.IsHold = packet.ReadBool();
         if (record.IsHold) {
             record.HoldInt = packet.ReadInt();
             record.HoldString = packet.ReadUnicodeString();
         }
 
-        session.Player.InBattle = true;
+        if (itemUid > 0) {
+            Item? item = session.Item.Inventory.Get(itemUid);
```
kOchirasu, line 116: Shouldn't normally happen, but we should protect against `Consume` failing.

```cs
Item? item = session.Item.Inventory.Get(itemUid);
// TODO: Check if item is valid for skill?
if (item == null || !session.Item.Inventory.Consume(item.Uid, 1)) {
    session.Send(NoticePacket.Notice(NoticePacket.Flags.Alert, StringCode.s_err_invalid_item));
    return;
}
```

Maple2.Server.Game/Util/SkillUtils.cs
```diff
@@ -119,4 +139,13 @@ public static class SkillUtils {
 
         return true;
     }
+
+    private static bool Check(this BeginConditionWeapon weapon, FieldPlayer player) {
+        bool IsValid(int handType, EquipSlot slot) {
```
kOchirasu, line 144: nit: can you move this function below the return statement (also with a line break between) so it's obvious that this is a function and not code being run when the function is called.

Maple2.Model/Metadata/BeginCondition.cs
Zintixx: You sort of have it laid out in `ItemType` record struct. I believe the lua files use these values too.
We can definitely do it for equips/weapons but everything else I'm not sure about (e.g. ObjectWeapon can be a value between 80 and 89). There is also the case where furnishing blocks start to collide with other type #s.
We can do a `EquipType` enum if that works?

kOchirasu: Would it make sense to just use that struct here instead? (With `Group` hardcoded as 1)

Maple2.Server.Game/Manager/Config/BuffManager.cs
kOchirasu: I meant whether or not we need to send the packet to remove the buffs instead of just clearing them. Also this function doesn't look like a strict "RemoveAll" anymore.

Zintixx: oh! no we don't need to send packets for it. I was mainly considering using this for the job change command but yeah that didn't make sense. I removed it and put the proper calls on the command instead.

kOchirasu: btw `Buffs` is public so if you're just going to use this for a command you can probably just remove this function entirely and avoid any confusion.

Maple2.File.Ingest/Mapper/AdditionalEffectMapper.cs
```diff
@@ -72,11 +80,31 @@ public class AdditionalEffectMapper : TypeMapper<AdditionalEffectMetadata> {
...
+    private static AdditionalEffectMetadataReflect Convert(ReflectProperty reflect) {
+        var values = new Dictionary<BasicAttribute, long>();
+        var rates = new Dictionary<BasicAttribute, float>();
+        
+        values.AddIfNotDefault(BasicAttribute.PhysicalAtk, reflect.physicalReflectionValue);
+        values.AddIfNotDefault(BasicAttribute.MagicalAtk, reflect.magicalReflectionValue);
+        
+        rates.AddIfNotDefault(BasicAttribute.PhysicalAtk, reflect.physicalReflectionRate);
+        rates.AddIfNotDefault(BasicAttribute.MagicalAtk, reflect.magicalReflectionRate);
+        return new AdditionalEffectMetadataReflect(
```
kOchirasu, line 97: nit: use named parameters in the constructor.

Maple2.Model/Metadata/AdditionalEffectMetadata.cs
```diff
@@ -18,31 +19,48 @@ public record AdditionalEffectMetadata(
...
 
 public record AdditionalEffectMetadataConsume(
     float HpRate,
     float SpRate);
 
+public record AdditionalEffectMetadataReflect(
+    float Rate,
+    int EffectId,
+    short EffectLevel,
+    int ReflectionCount,
```
kOchirasu, line 50: nit: "Reflect" is redundant here. Rename to things like "PhysicalRateLimit"

Maple2.Server.Game/Model/Skill/ReflectRecord.cs
```diff
@@ -0,0 +1,31 @@
...
+    public readonly int MaxCount;
+    public int Counter = 0;
+    public readonly int PhysicalRateLimit;
+    public readonly int MagicalRateLimit;
+    public IReadOnlyDictionary<BasicAttribute, long> ReflectValues;
+    public IReadOnlyDictionary<BasicAttribute, float> ReflectRates;
+    
+    public ReflectRecord(int id, AdditionalEffectMetadataReflect reflect) {
+        SourceBuffId = id;
+        Rate = reflect.Rate;
```
kOchirasu, line 22: Why copy all these values rather than storing metadata directly. It's readonly so there's no downside.

Maple2.Server.Game/Model/Field/Actor/Actor.cs
```diff
@@ -83,6 +80,17 @@ public abstract class Actor<T> : IActor<T>, IDisposable {
         }
     }
 
+    public virtual void Reflect(IActor target) {
+        if (Buffs.Reflect == null || Buffs.Reflect.Counter >= Buffs.Reflect.MaxCount) {
+            Console.WriteLine("No reflect buff found");
+            return;
+        }
+        Buffs.Reflect.Counter++;
```
kOchirasu, line 88: should you clear the reflect buff once it exceeds `MaxCount`?

Maple2.Server.Game/Model/Field/Buff.cs
```diff
@@ -77,6 +76,16 @@ public class Buff : IUpdatable, IByteSerializable {
         }
     }
 
+    public void Reflect() {
```
kOchirasu, line 79: feels very strange to have this indirection here. Why can't you just check this in `AddBuff` of `BuffManager`

PR #43, opened 2023-08-20
Achievement Manager

Zintixx:
- Adds a trophy manager ~and support for Account-wide trophies.~
  - ~I don't really like how I had to implement the trophy dictionary on accounts and how it's saved after. Looking for suggestions.~
- Added 2 new fields on CharacterUnlock used for trophies.
- Changed emotes in CharacterUnlock to be a List instead. This is because using a sorted set doesn't correctly display the ordering in game of emotes.
- Added a few more emotes upon character creation. Many of them are locked behind trophies so the player can't use it anyways until the trophy has been added.


Maple2.Database/Storage/Game/GameStorage.User.cs
```diff
@@ -68,16 +86,21 @@ public partial class Request : IPlayerInfoProvider {
...
-                    join account in Context.Account on character.AccountId equals account.Id
-                    join indoor in Context.UgcMap on
-                        new {OwnerId = character.AccountId, Indoor = true} equals new {indoor.OwnerId, indoor.Indoor}
-                    join outdoor in Context.UgcMap on
-                        new {OwnerId = character.AccountId, Indoor = false} equals new {outdoor.OwnerId, outdoor.Indoor} into plot
-                    from outdoor in plot.DefaultIfEmpty()
-                    select BuildPlayerInfo(character, indoor, outdoor, account.Trophy))
-                .FirstOrDefault();
+        public PlayerInfo? GetPlayerInfo(long characterId) 
+        {
```
kOchirasu, line 90: nit: brace

Maple2.Database/Storage/Metadata/AchievementMetadataStorage.cs
```diff
@@ -0,0 +1,48 @@
...
+                return false;
+            }
+
+            Cache.AddReplace(id, achievement);
+        }
+
+        return true;
+    }
+
+    public IEnumerable<AchievementMetadata> GetMany(AchievementConditionType type) {
```
kOchirasu, line 42: return `ICollection` here so you can get `.Count` if needed.

Maple2.File.Ingest/Mapper/AchievementMapper.cs
```diff
@@ -0,0 +1,103 @@
...
+                        Rank: grade.reward.rank)));
+            }
+
+            AchievementCategory category = AchievementCategory.Life;
+            string[] tags = data.categoryTag;
+            if (data.categoryTag.Length > 0) {
+                // skip the first in the array and use it as the trophy category
+                category = GetTrophyCategory(tags[0]);
+                tags = tags.Skip(1).ToArray();
+
```
kOchirasu, line 40: nit: extra line

Maple2.File.Ingest/Mapper/AchievementMapper.cs
```diff
@@ -0,0 +1,103 @@
...
+        if (codes.Length == 0) {
+            return null;
+        }
+        if (codes.Length > 1) {
+            List<int> integers = new();
+            List<string> strings = new();
+            foreach (string code in codes) {
+                if (!int.TryParse(code, out int intCode)) {
+                    strings.Add(code);
+                }
```
kOchirasu, line 72: nit: brackets

Maple2.Server.Game/Manager/AchievementManager.cs
```diff
@@ -0,0 +1,301 @@
...
+using Maple2.Tools.Extensions;
+using Serilog;
+
+namespace Maple2.Server.Game.Manager;
+
+public sealed class AchievementManager {
+    private const int BATCH_SIZE = 60;
+    private readonly GameSession session;
+
+    public IDictionary<int, Achievement> Values => session.Player.Value.Character.Achievements;
```
kOchirasu, line 21: shouldn't you keep 2 dictionaries here? (1 for Account, 1 for Character)

Maple2.Model/Metadata/AchievementMetadata.cs
```diff
@@ -0,0 +1,37 @@
...
+    AchievementMetadataCondition Condition,
+    AchievementMetadataReward? Reward);
+
+public record AchievementMetadataCondition(
+    AchievementConditionType Type,
+    AchievementMetadataCondition.Parameters? Codes,
+    long Value,
+    AchievementMetadataCondition.Parameters? Target) {
+
+    public record Parameters(
```
kOchirasu, line 25: what about setting all of these with default values of null so you don't have to pass null every time. (you only set 1 of the values anyways)

```cs
public record Paramters(string[]? Strings = null, Range<int>? Range = null, int[]? Integers = null);
```

Maple2.Server.Game/Packets/GuildPacket.cs
```diff
@@ -706,7 +706,7 @@ private enum Command : byte {
             pWriter.WriteLong(application.Guild.Id);
             pWriter.WriteUnicodeString(application.Guild.Name);
             pWriter.WriteUnicodeString(application.Guild.Emblem);
-            pWriter.Write<Trophy>(application.Guild.Trophy);
+            pWriter.WriteClass<Maple2.Model.Game.AchievementInfo>(application.Guild.AchievementInfo);
```
kOchirasu, line 709: why do we need the whole `Maple2.Model.Game.AchievementInfo` here? can we not import it?

Maple2.Server.Game/Util/Sync/PlayerInfoStorage.cs
```diff
@@ -45,7 +45,7 @@ public class PlayerInfoStorage {
                 Channel = (short) response.Channel,
                 UpdateTime = response.UpdateTime,
             };
-            var trophy = new Trophy {
+            var trophy = new Maple2.Model.Game.AchievementInfo {
```
kOchirasu, line 48: same as above, can we import `AchievementInfo`?

Maple2.Database/Model/Character.cs
```diff
@@ -101,6 +102,7 @@ internal class Character {
             StorageCooldown = other.Cooldown.Storage,
             DeleteTime = other.DeleteTime.ToEpochSeconds(),
             Channel = other.Channel,
+            Achievements = new Dictionary<int, Maple2.Model.Game.Achievement>(),
```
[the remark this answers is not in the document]
Zintixx: The reason is because it avoids doubling up fetching achievements twice.
I need to get the character's trophy count for `CharacterListPacket` when the user has logged in.
I would need to get it again when the player logs into the player.
I fetch it once this way.

Maple2.Database/Storage/Game/GameStorage.Achievement.cs
```diff
@@ -0,0 +1,87 @@
...
+                    case AchievementCategory.None:
+                    case AchievementCategory.Life:
+                        info.Lifestyle += trophy.Grades.Count;
+                        break;
+                }
+            }
+            return info;
+        }
+        
+        public bool SaveAchievements(long accountId, long characterId, IList<Achievement> values) {
```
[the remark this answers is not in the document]
Zintixx: the ownerId is dependent upon the trophy's metadata `AccountWide` bool. We don't know whether it's using accountId or characterId until we check that.

Maple2.Database/Model/Character.cs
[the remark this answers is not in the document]
kOchirasu: You would have to fetch twice anyway right? Once on `LoginServer` and once on `GameServer`.

I think a potential solution to this is to just store the Trophy counts for the character (used for display only, not as source of truth), but once you are logged in it will use values from `AchievementManager`.

Alternatively, it may be possible to craft a SQL query that can return trophy counts without needing to return all the achievement rows.

something like:
```sql
Count = SELECT SUM(GradeCount) FROM achievements WHERE OwnerId=AccountId OR OwnerId=CharacterId;
```

note: you would probably want to store a value for each row equivalent to `Grades.Count`

Maple2.Database/Storage/Game/GameStorage.Achievement.cs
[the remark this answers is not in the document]
kOchirasu: If you store two separate lists of Achievements, you would know that each one only contains Account or Character achievements.

Maple2.Database/Storage/Game/GameStorage.Club.cs
```diff
@@ -43,17 +44,21 @@ public partial class Request {
...
+                          join account in Context.Account on member.Character.AccountId equals account.Id
+                          join indoor in Context.UgcMap on
+                              new {OwnerId=member.Character.AccountId, Indoor=true} equals new {indoor.OwnerId, indoor.Indoor}
+                          join outdoor in Context.UgcMap on
+                              new {OwnerId=member.Character.AccountId, Indoor=true} equals new {outdoor.OwnerId, outdoor.Indoor} into plot
+                          from outdoor in plot.DefaultIfEmpty()
+                          select new {account, member, indoor, outdoor})
+                .ToList();
+
+            return (from entry in result
```
kOchirasu, line 57: very strange syntax, can we just do:
```cs
return results.Select(result => {
    AchievementInfo achievementInfo = GetAchievementInfo(result.account.Id, result.member.CharacterId);
    PlayerInfo playerInfo = BuildPlayerInfo(result.member.Character, result.indoor, result.outdoor, achievementInfo);
    return new ClubMember(playerInfo, result.member.CreationTime.ToEpochSeconds(), result.member.Character.LastModified.ToEpochSeconds());
}).ToList();
```

Maple2.Database/Storage/Game/GameStorage.Club.cs
```diff
@@ -43,17 +44,21 @@ public partial class Request {
...
-                        member.Character.LastModified.ToEpochSeconds())).ToList();
+            var result = (from member in Context.ClubMember where member.ClubId == clubId
+                          join account in Context.Account on member.Character.AccountId equals account.Id
+                          join indoor in Context.UgcMap on
+                              new {OwnerId=member.Character.AccountId, Indoor=true} equals new {indoor.OwnerId, indoor.Indoor}
+                          join outdoor in Context.UgcMap on
+                              new {OwnerId=member.Character.AccountId, Indoor=true} equals new {outdoor.OwnerId, outdoor.Indoor} into plot
+                          from outdoor in plot.DefaultIfEmpty()
+                          select new {account, member, indoor, outdoor})
+                .ToList();
```
kOchirasu, line 55: curious whether or not we even need `ToList()` here since we are gonna enumerate it again.

Zintixx: It looks like it won't take it without it being a list.

kOchirasu: what about `AsEnumerable` which is a soft cast from `IQueryable` rather than needing to allocate a list.

Maple2.Database/Storage/Game/GameStorage.User.cs
```diff
@@ -48,18 +48,32 @@ public partial class Request : IPlayerInfoProvider {
...
         }
 
         //  If accountId is specified, only characters for the account will be returned.
         public Character? GetCharacter(long characterId, long accountId = -1) {
             if (accountId < 0) {
-                return Context.Character.Find(characterId);
+                Character? characterFind = Context.Character.Find(characterId);
+                if (characterFind != null) {
+                    characterFind.AchievementInfo = GetAchievementInfo(accountId, characterId);
+                }
```
kOchirasu, line 67: missing `return characterFind`

PR #45, opened 2023-08-23
More Trophy Types

Zintixx:
- Implements a handful more trophy types
- This also rewrites some functions on AchievementManager due to some edge cases, like grade 1 not always being available.
- This change also changes when the Achievement Update packet should be sent and shouldn't (to avoid redundancy)

### DIFF
--- a/Maple2.Model/Common/Color.cs
+++ b/Maple2.Model/Common/Color.cs
@@ -45,16 +45,6 @@ public readonly struct EquipColor {
         this.Index = -1;
         this.PaletteId = 0;
     }
-    
-    public bool Equals(EquipColor other) {
-        return Primary.Equals(other.Primary) && Secondary.Equals(other.Secondary) && Tertiary.Equals(other.Tertiary) && Index == other.Index && PaletteId == other.PaletteId;
-    }
-    public override bool Equals(object? obj) {
-        return obj is EquipColor other && Equals(other);
-    }
-    public override int GetHashCode() {
-        return HashCode.Combine(Primary, Secondary, Tertiary, Index, PaletteId);
-    }
 
     [JsonConstructor]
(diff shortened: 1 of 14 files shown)


Maple2.Model/Common/Color.cs
```diff
@@ -44,6 +45,16 @@ namespace Maple2.Model.Common;
         this.Index = -1;
         this.PaletteId = 0;
     }
+    
+    public bool Equals(EquipColor other) {
```
kOchirasu, line 49: do we need these? struct equality is already the same by default.

Maple2.Model/Common/Color.cs
```diff
@@ -54,6 +55,14 @@ namespace Maple2.Model.Common;
         this.Index = index;
     }
 
+    public static bool operator ==(EquipColor color1, EquipColor color2) {
```
kOchirasu, line 58: remove these too (and the using statements)

Maple2.Server.Game/PacketHandlers/BeautyHandler.cs
```diff
@@ -220,8 +222,14 @@ private enum Command : byte {
                 return;
             }
         }
-
-        ModifyBeauty(session, packet, session.BeautyShop.ShopType, cosmetic.Id);
+        
+        EquipColor? startColor = cosmetic.Appearance?.Color;
+        if (ModifyBeauty(session, packet, session.BeautyShop.ShopType, cosmetic.Id) && startColor != null) {
+            Item newCosmetic = session.Item.Equips.Get(cosmetic.Metadata.SlotNames.First())!;
+            if (newCosmetic.Appearance != null && !Equals(newCosmetic.Appearance?.Color, startColor)) {
```
kOchirasu, line 229: Isn't this enough:
```cs
if (!Equals(newCosmetic.Appearance?.Color, startColor)) {
```

PR #46, opened 2023-08-23
GemSockets Slots, Constant Stats, and Static stats

Zintixx:
- Implements using lua script for gem socket slot count. Also has a quick open socket count formula. We're assuming the success rate is 5%.
- Implements using Constant and Static Stats for items defined in the tables.

### DIFF
--- a/Maple2.Server.Game/Util/ItemStatsCalculator.cs
+++ b/Maple2.Server.Game/Util/ItemStatsCalculator.cs
@@ -150,7 +150,7 @@ public sealed class ItemStatsCalculator {
 
         if (item.Metadata.Option.LevelFactor > 50 && pick != null) {
             if (option == null) {
-                option = new ItemStats.Option(new Dictionary<BasicAttribute, BasicOption>(), new Dictionary<SpecialAttribute, SpecialOption>());
+                option = new ItemStats.Option();
             }
 
             foreach ((BasicAttribute attribute, int deviation) in pick.ConstantValue) {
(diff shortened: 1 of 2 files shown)


Maple2.Server.Game/Util/ItemStatsCalculator.cs
```diff
@@ -157,6 +137,64 @@ public sealed class ItemStatsCalculator {
         return true;
     }
 
+    // Used to calculate the default constant attributes for a given item.
+    private bool GetConstantOption(Item item, int job, ItemOptionPickTable.Option? pick, [NotNullWhen(true)] out ItemStats.Option? option) {
```
kOchirasu, line 141: can you explain this function, I'm not able to deduce its rules

Zintixx: The function will get all the stats defined in the ItemOptionConstantTable first.

Constant stats only use the ItemOptionPick table if their LevelFactor is over 50.
IF there were no stats in the Constant Table, it'll just run the Pick Table and lua functions as is. However, if there was a table, it'll feed the stat value into the lua function. It modifies the value a bit. I don't think ALL the lua functions make use of the stat value.

Maple2.Server.Game/Util/ItemStatsCalculator.cs
```diff
@@ -157,6 +139,64 @@ public sealed class ItemStatsCalculator {
...
+        }
+
+        if (TableMetadata.ItemOptionStaticTable.Options.TryGetValue(item.Metadata.Option.StaticId, item.Rarity, out ItemOption? itemOption)) {
+            // We're using RandomItemOption here considering the logic is the same.
+            option = RandomItemOption(itemOption);
+        }
+
+        if (pick != null) {
+            if (option == null) {
+                option = new ItemStats.Option(new Dictionary<BasicAttribute, BasicOption>(), new Dictionary<SpecialAttribute, SpecialOption>());
```
kOchirasu, line 185: default constructor here will create the dictionaries for you.

PR #47, opened 2023-08-23
UGC Web Server

Zintixx:
Big thanks to @AngeloTadeucci from the other emulator 🙂

### DIFF
--- a/Maple2.Model/Enum/Ugc.cs
+++ b/Maple2.Model/Enum/Ugc.cs
@@ -1,0 +1,19 @@
+﻿namespace Maple2.Model.Enum;
+
+public enum UgcType : byte
+{
+    None = 0,
+    Item = 1,
+    Furniture = 2,
+    Banner = 3,
+    Unknown4 = 4,
+    ProfileAvatar = 5,
+    GuildEmblem = 6,
+    Mount = 7,
+    GuildBanner = 8,
+    Unknown9 = 9,
+    Unknown10 = 10,
+    ItemIcon = 201,
+    Unknown11 = 202,
+    Unknown12 = 209,
+}
--- a/Maple2.Model/Enum/Ugc.cs
+++ b/Maple2.Model/Enum/Ugc.cs
@@ -1,7 +1,6 @@
 ﻿namespace Maple2.Model.Enum;
 
-public enum UgcType : byte
-{
+public enum UgcType : byte {
     None = 0,
     Item = 1,
     Furniture = 2,
--- a/Maple2.Model/Enum/Ugc.cs
+++ b/Maple2.Model/Enum/Ugc.cs
@@ -1,6 +1,6 @@
 ﻿namespace Maple2.Model.Enum;
 
-public enum UgcType : byte {
+public enum UgcType {
     None = 0,
     Item = 1,
     Furniture = 2,
--- a/Maple2.Server.Login/PacketHandlers/UgcHandler.cs
+++ b/Maple2.Server.Login/PacketHandlers/UgcHandler.cs
@@ -32,8 +32,6 @@ public class UgcHandler : PacketHandler<LoginSession> {
             case Command.ProfilePicture:
                 HandleProfilePicture(session, packet);
                 break;
-            default:
-                return;
         }
     }
 
--- a/Maple2.Server.Login/PacketHandlers/UgcHandler.cs
+++ b/Maple2.Server.Login/PacketHandlers/UgcHandler.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using Grpc.Core;
+using Maple2.Database.Storage;
+using Maple2.Model.Game;
+using Maple2.PacketLib.Tools;
+using Maple2.Server.Core.Constants;
+using Maple2.Server.Core.PacketHandlers;
+using Maple2.Server.Core.Packets;
+using Maple2.Server.Login.Session;
+using Maple2.Server.World.Service;
+using static Maple2.Model.Error.MigrationError;
+using WorldClient = Maple2.Server.World.Service.World.WorldClient;
+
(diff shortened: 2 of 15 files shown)


Maple2.Server.Web/Endpoints/Upload.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        byte[]? fileBytes = packet.ReadBytes(packet.Available);
+
+        return type switch {
+            UgcType.ProfileAvatar => HandleProfileAvatar(fileBytes, characterId),
+            _ => HandleUnknownMode(type)
+        };
+    }
+
+    private static IResult HandleProfileAvatar(byte[] fileBytes, long characterId) {
```
kOchirasu, line 37: probably okay for now, but can we switch to storing this in the database?

Maple2.Server.Web/Endpoints/Upload.cs
```diff
@@ -0,0 +1,68 @@
...
+        string filePath = $"{Target.DataDir}/profiles/{characterId}/";
+        Directory.CreateDirectory(filePath);
+
+        string uniqueFileName = Guid.NewGuid().ToString();
+
+        // Deleting old files in the character folder
+        DirectoryInfo di = new(filePath);
```
kOchirasu, line 44: no short variable names like this unless it's an obvious acronym (e.g. id). Please use `directory` or something

Maple2.Server.Web/Endpoints/Upload.cs
```diff
@@ -0,0 +1,68 @@
...
+        File.WriteAllBytes($"{filePath}/{uniqueFileName}.png", fileBytes);
+        return Results.Text($"0,data/profiles/avatar/{characterId}/{uniqueFileName}.png");
+    }
+
+    private static IResult HandleUnknownMode(UgcType mode) {
+        Log.Logger.Warning("Unknown upload mode: {mode}", mode);
+        return Results.BadRequest();
+    }
+
+    private static async Task<MemoryStream> CopyStream(Stream input) {
```
kOchirasu, line 58: does [Stream.CopyTo](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.copyto?view=net-7.0) work

Zintixx: CopyToAsync does. Pushed a change for this.

Maple2.Server.Login/PacketHandlers/UgcHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        ProfilePicture = 11,
+    }
+    
+    public override void Handle(LoginSession session, IByteReader packet) {
+        var command = packet.Read<Command>();
+        switch (command) {
+            case Command.ProfilePicture:
+                HandleProfilePicture(session, packet);
+                break;
+            default:
```
kOchirasu, line 35: default case is useless

Maple2.Model/Enum/Ugc.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Maple2.Model.Enum;
+
+public enum UgcType : byte {
```
kOchirasu, line 3: in the only place I see this used, it's used as an `int`, is there another reason this is `byte`? `int` will allow you to do `Read<UgcType>()`

Maple2.Server.Web/Endpoints/Upload.cs
```diff
@@ -0,0 +1,64 @@
...
+using Serilog;
+
+namespace Maple2.Server.Web.Endpoints;
+
+public static class UploadEndpoint {
+    public static async Task<IResult> Post(HttpRequest request) {
+        Stream bodyStream = request.Body;
+
+        MemoryStream memoryStream = await CopyStream(bodyStream);
+        if (memoryStream.Length == 0) {
```
kOchirasu, line 14: why not check `bodyStream.Length` directly?

Maple2.Server.Web/Endpoints/Upload.cs
kOchirasu: i would just get rid of this function and move the code inline (i also think the async is useless here since you await immediately, unless it's mandatory? async stuff is weird)

Maple2.Server.Web/Endpoints/Upload.cs
Zintixx: yeah it's strange. I get:
`Failed to read field: Value does not fall the expected range.`
i've tried reading length on request.Body too.

Maple2.Model/Enum/Ugc.cs
Zintixx: I think it should be fine.

Maple2.Server.Web/Endpoints/Upload.cs
Zintixx: Yeah it appears to be mandatory due to the function being async

Maple2.Server.Web/Maple2.Server.Web.csproj
```diff
@@ -0,0 +1,23 @@
...
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+        <LangVersion>11</LangVersion>
+        <RootNamespace>Maple2.Server.Web</RootNamespace>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Maple2.PacketLib" Version="1.0.2"/>
+        <PackageReference Include="Serilog.AspNetCore" Version="7.0.0"/>
+        <PackageReference Include="Serilog.Expressions" Version="3.4.2-dev-00120"/>
```
kOchirasu, line 15: is using dev release necessary?

PR #50, opened 2023-08-26
Exp on base actions/processes

Zintixx:
- This is using an updated Maple2.File.Parser and shouldn't be merged until it's updated
- Adds and parses CommonExp Table
  - Currently implements a handful of giving exp on features like fishing, music, taxi, etc.
- Add a simple exp command
- Implement level/level up trophies. 44 more finished!


Maple2.Server.Game/Manager/ExperienceManager.cs
```diff
@@ -57,18 +60,76 @@ public sealed class ExperienceManager {
...
         RestExp = Math.Max(0, RestExp - addedRestExp);
         Exp += expGained;
         return addedRestExp;
     }
 
     public void AddExp(long expGained, ExpMessageCode expMessageCode = ExpMessageCode.s_msg_take_exp) {
         expGained += GetRestExp(expGained);
-        LevelUp();
+        if (LevelUp()) {
+            session.PlayerInfo.SendUpdate(new PlayerUpdateRequest {
```
kOchirasu, line 72: why not move this into the `LevelUp()` function since it's more related to `LevelUp` than `AddExp`?

Maple2.Server.Game/PacketHandlers/InstrumentHandler.cs
```diff
@@ -267,6 +285,7 @@ private enum Command : byte {
         }
 
         fieldInstrument = session.Field!.SpawnInstrument(session.Player, metadata);
+        fieldInstrument.StartTick = Environment.TickCount;
```
kOchirasu, line 288: For any tick values that get stored, I've tried to use `TickCount64()` instead. `TickCount()` will wrap around after 25 days system uptime and could cause problems. It's fine to cast the value back to `int` if you need to use it as an `int` since that's equivalent to being truncated, but in this case since you're just finding the delta it won't be a problem.

PR #52, opened 2023-08-28
Meret Market Pt1: Premium Shop

Zintixx:


Maple2.Database/Storage/Game/GameStorage.Market.cs
```diff
@@ -54,5 +59,109 @@ public partial class Request {
...
+                                        entry.TabId == tabId)
+                        .ToList()!;
+                    break;
+            }
+
+            foreach (MarketItem marketEntry in results) {
+                if (marketEntry is not PremiumMarketItem premium) {
+                    continue;
+                }
+                premium.AdditionalQuantities = Context.PremiumMarketItem
```
kOchirasu, line 141: Very bad, this seems like a full table scan.
`Context.PremiumMarketItem.Select(ToMarketEntry)`

Maple2.Database/Storage/Game/GameStorage.SystemBanner.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Linq;
+using Maple2.Model.Game;
+
+namespace Maple2.Database.Storage;
+
+public partial class GameStorage {
+    public partial class Request {
+        public IList<SystemBanner> GetBanners() {
+            return Context.SystemBanner
+                .AsEnumerable()
```
kOchirasu, line 12: Once you call `AsEnumerable()` the SQL query is finalized and all subsequent logic is executed locally. This means the whole table is going to be returned here. Surely it's possible to have the DB perform the time filter below right?

Maple2.Database/sql/system_banner.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO `game-server`.`system-banner` (Name,`Type`,`Function`,FunctionParameter,Url,`Language`,BeginTime,EndTime) VALUES
+	 ('homeproduct_emulator',1,1,'','https://i.imgur.com/TV47C2m.png',-1,'2010-01-01 00:00:00','2030-01-01 00:00:00');
```
kOchirasu, line 2: how about hosting this image on `Maple2.Server.Web`?

Maple2.File.Ingest/Mapper/TableMapper.cs
```diff
@@ -1111,6 +1114,46 @@ public class TableMapper : TypeMapper<TableMetadata> {
...
+                                {{tab.id, tabEntry}});
+                        } else {
+                            (results[id] as Dictionary<int, MeretMarketCategoryTable.Tab>)!.Add(tab.id, tabEntry);
+                        }
+                    }
+                }
+            }
+        }
+        return new MeretMarketCategoryTable(results);
+	}
```
kOchirasu, line 1155: nit: brace indentation

Maple2.File.Ingest/Mapper/TableMapper.cs
```diff
@@ -1111,6 +1114,46 @@ public class TableMapper : TypeMapper<TableMetadata> {
         return new Color(color.B, color.G, color.R, color.A);
     }
 
+    private MeretMarketCategoryTable ParseMeretMarketCategoryTable() {
+        var results = new Dictionary<int, IReadOnlyDictionary<int, MeretMarketCategoryTable.Tab>>();
+        if (FeatureLocaleFilter.FeatureEnabled("MeratMarketNewBM")) {
```
kOchirasu, line 1119: let's invert this if statement to reduce nesting:
```cs
if (!FeatureLocaleFilter.FeatureEnabled("MeratMarketNewBM")) {
    return ...;
}

File: Maple2.Model/Enum/Gender.cs [hunk 6]
```diff
@@ -1,7 +1,15 @@
...
+namespace Maple2.Model.Enum;
 
 public enum Gender : byte {
     Male = 0,
     Female = 1,
     All = 2,
 }
+
+[Flags]
+public enum GenderFlag : byte {
```
kOchirasu, line 12: To make it more clear that this is specifically for market filtering, I would move to another file and name it `GenderFilterFlag` or something. Same with `JobFlag` below. (`MeretMarket.cs` seems like a good choice, unless there are other places the enum is used)

Maple2.Model/Game/Market/MarketItem.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Maple2.Model.Metadata;
+using Maple2.PacketLib.Tools;
+using Maple2.Tools;
+
+namespace Maple2.Model.Game;
+
+public abstract class MarketItem : IByteSerializable {
+    public readonly ItemMetadata ItemMetadata;
+    public int ItemId { get; init; }
```
kOchirasu, line 9: `ItemId` should exist in `ItemMetadata`

Maple2.Model/Game/Market/PremiumMarketItem.cs
```diff
@@ -0,0 +1,86 @@
+﻿using System.Collections.Generic;
+using Maple2.Model.Enum;
+using Maple2.Model.Metadata;
+using Maple2.PacketLib.Tools;
+
+namespace Maple2.Model.Game;
+
+public class PremiumMarketItem : MarketItem {
+    public int Id { get; init; }
```
kOchirasu, line 9: everywhere you're using `init`, you can also use `required` to ensure values are set when constructing the object.
```cs
public required int Id { get; init; }
```

Maple2.Model/Metadata/Table/MeretMarketCategoryTable.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections.Generic;
+using Maple2.Model.Common;
+
+namespace Maple2.Model.Metadata;
+
+public record MeretMarketCategoryTable(IReadOnlyDictionary<int, IReadOnlyDictionary<int, MeretMarketCategoryTable.Tab>> Entries) : Table {
+
+    public record Tab(List<string> Categories,
```
kOchirasu, line 8: Use arrays instead of lists for Metadata.

Maple2.Server.Game/PacketHandlers/MeretMarketHandler.cs
```diff
@@ -0,0 +1,296 @@
...
+namespace Maple2.Server.Game.PacketHandlers;
+
+public class MeretMarketHandler : PacketHandler<GameSession> {
+    public override RecvOp OpCode => RecvOp.MeretMarket;
+
+    #region Autofac Autowired
+    // ReSharper disable MemberCanBePrivate.Global
+    public required ItemMetadataStorage ItemMetadata { private get; init; }
+    public required TableMetadataStorage TableMetadata { private get; init; }
+
```
kOchirasu, line 26: nit: empty line

Maple2.Server.World/Migrations/20230827232716_MeretMarket.cs
```diff
@@ -0,0 +1,130 @@
...
+                    CreationTime = table.Column<DateTime>(type: "datetime(6)", nullable: false)
+                        .Annotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.IdentityColumn)
+                },
+                constraints: table =>
+                {
+                    table.PrimaryKey("PK_ugc-market-item", x => x.Id);
+                })
+                .Annotation("MySql:CharSet", "utf8mb4");
+            
+            migrationBuilder.Sql("ALTER TABLE `ugc-market-item` AUTO_INCREMENT = 10000000");
```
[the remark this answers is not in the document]
Zintixx: IIRC, there were some funky issues starting the id off at 1. I only used this number to pad it just incase. It really could just start off at 1000.

Maple2.Database/sql/system_banner.sql
Zintixx: What do you have in mind? I wasn't thinking of putting it in ugc-resources because it's not ugc. Banners also require the full url

Maple2.Database/Storage/Game/GameStorage.Market.cs
Zintixx: I'm not sure what you mean? Isn't that what I'm doing? I would have to look through the entire table for it.

Maple2.Database/sql/system_banner.sql
kOchirasu: `Maple2.Server.Web` is essentially just a website, so you could host anything on it.

Maple2.Database/Storage/Game/GameStorage.Market.cs
kOchirasu: So the way SQL works is there is an engine that can execute queries on the DB server before returning results. Ideally you limit the amount of data returned to you because it's much more efficient to let the DB engine handle that part.

In this case, you would do something like:
```cs
Context.PremiumMarketItem
    .Where(subEntry => subEntry != null && subEntry.ParentId == premium.Id) // On DB
    .Select(ToMarketEntry) // Local
    .ToList()!; // Local
```

Zintixx: Makes sense. I pushed a few updates to remove some repeated code and fix this. hopefully.

Maple2.File.Ingest/Mapper/TableMapper.cs
```diff
@@ -1111,6 +1114,53 @@ public class TableMapper : TypeMapper<TableMetadata> {
...
 
+    private MeretMarketCategoryTable ParseMeretMarketCategoryTable() {
+        string feature = FeatureLocaleFilter.FeatureEnabled("MeratMarketNewBM") ? "MeratMarketNewBM" : string.Empty;
+        Dictionary<int, IReadOnlyDictionary<int, MeretMarketCategoryTable.Tab>> results = GetMereMarketTabs(feature);
+        return new MeretMarketCategoryTable(results);
+	}
+
+    private Dictionary<int, IReadOnlyDictionary<int, MeretMarketCategoryTable.Tab>> GetMereMarketTabs(string featureEnabled = "") {
+        var results = new Dictionary<int, IReadOnlyDictionary<int, MeretMarketCategoryTable.Tab>>();
+        foreach ((int id, string feature, MeretMarketCategory category) in parser.ParseMeretMarketCategory()) {
```
kOchirasu, line 1125: idk why feature is being returned here, but you **never** need to filter by feature manually. This is already done by the parsing framework.

There's a reason the variable name begins with `_`, it means don't use it.
`yield return (category.id, environment._feature, category);`

The problem seems to be `XmlElement` instead of `M2dFeatureLocale`: https://github.com/kOchirasu/Maple2.File/blob/443583860cdb2d9f5e85dd1d49facac208cb5d7e/Maple2.File.Parser/Xml/Table/MeretMarketCategory.cs#L11

Maple2.Server.Web/Controllers/SystemBannerController.cs
```diff
@@ -0,0 +1,23 @@
...
+[Route("/system/")]
+public class SystemBannerController : ControllerBase {
+    private static readonly string SolutionDir = Path.GetFullPath(Path.Combine(AppContext.BaseDirectory, "../../../.."));
+    private static readonly string RootDir = Path.Combine(SolutionDir, "Maple2.Server.Web/Data");
+
+    [HttpGet("banner/{name}.png")]
+    public IResult GetBanner(string name) {
+        string fullPath = $"{RootDir}/system/banner/{name}.png";
+        if (!System.IO.File.Exists(fullPath)) {
+            return Results.BadRequest();
```
kOchirasu, line 17: `Results.NotFound()` is more appropriate here.

Maple2.Server.Web/Controllers/SystemBannerController.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+using System.IO;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Maple2.Server.Web.Controllers;
+
+[Route("/system/")]
+public class SystemBannerController : ControllerBase {
```
kOchirasu, line 9: I would call this `SystemController` since that's the root route.

Maple2.File.Ingest/Mapper/TableMapper.cs
Zintixx: It's because the xml is written a bit different:
`	<environment feature="MeratMarketNewBM" locale="">
		<category id="100000" tabType="sort">
			<tab name="40300" id="40300" symbol="UIMeratMarketDialog_symbol_etc" priority="6" static="1" />
			<tab name="40600" id="40600" symbol="UIMeratMarketDialog_symbol_instrument" priority="9" static="1" />
		</category>
	</environment>`

Maple2.Database/Storage/Game/GameStorage.Market.cs
```diff
@@ -54,5 +58,59 @@ public partial class Request {
             Context.MesoMarket.Remove(listing);
             return Context.TrySaveChanges();
         }
+
+        public ICollection<PremiumMarketItem> GetPremiumMarketEntries(GenderFilterFlag gender, JobFilterFlag job, string searchString, params int[] tabIds) {
```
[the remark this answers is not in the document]
Zintixx: Were you suggesting caching it in GameSession.State? 
Then just refetching whenever startpage is 1/different tab

Or on the server level?

Maple2.Model/Game/Market/PremiumMarketItem.cs
```diff
@@ -0,0 +1,85 @@
+﻿using System.Collections.Generic;
+using Maple2.Model.Enum;
+using Maple2.Model.Metadata;
+using Maple2.PacketLib.Tools;
+
+namespace Maple2.Model.Game;
+
+public class PremiumMarketItem : MarketItem {
+    public required int Id { get; init; }
```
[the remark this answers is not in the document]
Zintixx: This id is the auto generated key, not item id.

Maple2.Database/Storage/Game/GameStorage.Market.cs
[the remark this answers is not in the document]
kOchirasu: Yea server level, but you can use your own judgement on how to store it. I don't know what operations we need to perform on the data. My instinct is that we cache the exact result from the db, since all post filtering can be derived from that.

Maple2.File.Ingest/Mapper/TableMapper.cs
kOchirasu: I don't understand what's different

Zintixx: It's wrapped in environment, while the "normal" one isn't.
for comparison:
https://github.com/Zintixx/MapleStory2-English/blob/master/Latest-GMS2/table/na/meratmarketcategory.xml

Zintixx: It would, but I would still need to do a check and ingest the correct section if the feature is enabled, wouldn't I?

kOchirasu: You should not because `environment` would be empty if the feature is not enabled.

Maple2.Server.Game/PacketHandlers/MeretMarketHandler.cs
```diff
@@ -0,0 +1,310 @@
...
+                session.Currency.Meret -= price;
+                return StringCode.s_empty_string;
+            case MeretMarketCurrencyType.RedMeret: // TODO: Implement Red Meret
+                return StringCode.s_err_lack_merat_red;
+            default:
+                return StringCode.s_err_lack_money;
+        }
+    }
+
+    private static ICollection<PremiumMarketItem> Filter(IEnumerable<PremiumMarketItem> items, GenderFilterFlag gender, JobFilterFlag job, string searchString) {
```
kOchirasu, line 267: I would return `IEnumerable` here so you don't have to allocate the `List`. Since these intermediate steps are chained and you only need the list at the very end.

Maple2.Server.Game/PacketHandlers/MeretMarketHandler.cs
```diff
@@ -0,0 +1,310 @@
...
+
+        // Featured page needs a multiple of 2 slots. Add to the entries total if odd number.
+        byte marketSlots = (byte) (entries.Count % 2 == 0 ? entries.Count : (byte) (entries.Count + 1));
+        session.Send(MeretMarketPacket.Featured(section, tabId, marketSlots, entries.ToList()));
+    }
+
+    private void HandleFindItem(GameSession session, IByteReader packet) {
+        bool premium = packet.ReadBool();
+        MarketItem? marketItem;
+        using GameStorage.Request db = session.GameStorage.Context();
```
kOchirasu, line 180: `db` is not used

Maple2.Server.Game/PacketHandlers/MeretMarketHandler.cs
```diff
@@ -0,0 +1,310 @@
...
+
+        session.Send(MeretMarketPacket.Purchase(totalQuantity, itemIndex, price, premiumMarketId, ugcItemId));
+    }
+
+    private static void HandleFeatured(GameSession session, IByteReader packet) {
+        byte section = packet.ReadByte();
+        byte tabId = packet.ReadByte();
+
+        ICollection<MarketItem> entries = new List<MarketItem>();
+        using GameStorage.Request db = session.GameStorage.Context();
```
kOchirasu, line 161: `db` is not used

Maple2.Server.Game/PacketHandlers/MeretMarketHandler.cs
```diff
@@ -0,0 +1,310 @@
...
+        price = entry.SalePrice > 0 ? entry.SalePrice : entry.Price;
+        StringCode payResult = Pay(session, price, entry.CurrencyType);
+        if (payResult != StringCode.s_empty_string) {
+            session.Send(NoticePacket.MessageBox(payResult));
+            return;
+        }
+
+        Item? newItem = session.Item.CreateItem(entry.ItemMetadata.Id, entry.Rarity, entry.Quantity + entry.BonusQuantity);
+        if (newItem == null) {
+            return;
```
kOchirasu, line 142: This should be a fatal error since the player already paid for it.

```cs
Logger.Fatal("Failed to create item {ItemId}, {Rarity}, {Quantity}", entry.ItemMetadata.Id, entry.Rarity, entry.Quantity + entry.BonusQuantity);
throw new InvalidOperationException($"Fatal: ...");
```

Maple2.File.Ingest/Mapper/TableMapper.cs
Zintixx: **If the feature wasn't enabled:**

It would not ingest environment and only parse the "normal" meret market.

**If the feature was enabled:**
It would ingest both. You would still need some kind of check whether it should ignore the "normal" meret market.

I understand this is how it should be properly parsed but it doesn't eliminate the need to check feature enabled when ingesting.

kOchirasu: We can also add the check in the parser itself:
```cs
if (environment.Count > 0) {
    // return environment results
}
// return normal results
```

Maple2.Database/Storage/Game/GameStorage.Market.cs
```diff
@@ -54,5 +56,35 @@ public partial class Request {
             Context.MesoMarket.Remove(listing);
             return Context.TrySaveChanges();
         }
+
+        public ICollection<PremiumMarketItem> GetMarketItems() {
+            IEnumerable<PremiumMarketItem> selectedResults = Context.PremiumMarketItem
+                .Where(entry => entry.ParentId == 0)
+                .AsEnumerable()
+                .Select(ToMarketEntry)
+                .ToList()!;
```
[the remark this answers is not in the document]
Zintixx: Without it I error:
`This MySqlConnection is already in use. `

Maple2.File.Ingest/Mapper/TableMapper.cs
Zintixx: Okay sure we can do that. Did you want to include this on your PR for Maple2.File or I could open one?

Maple2.Server.Game/PacketHandlers/MeretMarketHandler.cs
```diff
@@ -0,0 +1,310 @@
...
+        packet.ReadByte();
+        packet.ReadByte();
+        byte itemsPerPage = packet.ReadByte();
+        MeretMarketSection section = GetMarketSection(packet.ReadByte());
+
+        ICollection<MarketItem> entries = GetItems(session, section, sortBy, 0, gender, job, searchString);
+        int totalItems = entries.Count;
+        entries = TakeLimit(entries, startPage, itemsPerPage);
+
+        session.Send(MeretMarketPacket.LoadItems(entries.ToList(), totalItems, startPage));
```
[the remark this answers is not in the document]
Zintixx: I error if I remove ToList():

`Argument type 'System.Collections.Generic.ICollection<Maple2.Model.Game.MarketItem>' is not assignable to parameter type 'System.Collections.Generic.IList<Maple2.Model.Game.MarketItem>'`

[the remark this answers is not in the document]
kOchirasu: change the parameter of LoadItems to `ICollection`

Maple2.Database/Storage/Game/GameStorage.Market.cs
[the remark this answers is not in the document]
kOchirasu: Ok, then don't call `ToList()` again in `GetAdditionalQuantaties`:
```cs
private void SetAdditionalQuantities(IList<PremiumMarketItem> selectedItems) {
    foreach (PremiumMarketItem marketEntry in selectedItems) {
        marketEntry.AdditionalQuantities = Context.PremiumMarketItem
            .Where(subEntry => subEntry.ParentId == marketEntry.Id)
            .AsEnumerable()
            .Select(ToMarketEntry)
            .ToList()!;
    }
}
```

Maple2.Server.Game/PacketHandlers/MeretMarketHandler.cs
```diff
@@ -0,0 +1,310 @@
...
+                }
+                session.Currency.Meso -= price;
+                return StringCode.s_empty_string;
+            case MeretMarketCurrencyType.Meret:
+                if (session.Currency.CanAddMeret(-price) != -price) {
+                    return StringCode.s_err_lack_merat;
+                }
+                session.Currency.Meret -= price;
+                return StringCode.s_empty_string;
+            case MeretMarketCurrencyType.RedMeret: // TODO: Implement Red Meret
```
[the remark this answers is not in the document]
Zintixx: RedMeret isn't the same as GameMeret, no?
It even has the Red Meret icon:
https://i.imgur.com/G5vFI2z.png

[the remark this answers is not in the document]
kOchirasu: `CurrencyManager` implies that it's the same: https://github.com/kOchirasu/Maple2/blob/d505ce2b6dd24993525da7741edb499c5c8cd21a/Maple2.Server.Game/Manager/CurrencyManager.cs#L36

Maple2.Database/Storage/Game/GameStorage.Market.cs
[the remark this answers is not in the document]
kOchirasu: Actually, I would probably just get rid of this function, a function with side-effects is bad style. Merge the code into the other function.

Maple2.File.Ingest/Mapper/TableMapper.cs
kOchirasu: I'll make the change since I have it already

kOchirasu: resolved

PR #54, opened 2023-09-11
Ugc Items & Ugc Market

Zintixx:
So unlike Premium Market, I'm not caching this because there's a lot of moving parts (sales count, unlisting, relisting, etc)


Maple2.Database/Model/UgcResource.cs
```diff
@@ -17,13 +19,22 @@ internal class UgcResource {
...
+            Id = other.Id,
+            Path = other.Path,
+            Type = other.Type,
         };
     }
 
     public static void Configure(EntityTypeBuilder<UgcResource> builder) {
         builder.ToTable("ugcresource");
         builder.HasKey(ugc => ugc.Id);
-        builder.HasIndex(ugc => ugc.OwnerId);
```
kOchirasu, line 26: why remove this index, it seems useful.

Maple2.Database/Storage/Game/GameStorage.Market.cs
```diff
@@ -74,12 +163,72 @@ public partial class Request {
...
+                item.Category = UgcMarketHomeCategory.Promoted;
+            }
+            return items;
+        }
+
+        public ICollection<UgcMarketItem> GetUgcMarketNewItems() {
+            ICollection<UgcMarketItem> items = Context.UgcMarketItem
+                .Where(item => item.ListingEndTime > DateTime.Now)
+                .AsEnumerable()
+                .OrderBy(item => item.CreationTime)
```
kOchirasu, line 201: can we do this before `AsEnumerable()`, the SQL engine should be able to evaluate it with `ORDER BY` and `LIMIT` limiting the number of entries that need to be returned from db.

Maple2.Model/Game/Market/UgcMarketItem.cs
```diff
@@ -6,27 +6,31 @@
 
 namespace Maple2.Model.Game;
 
-public class UgcMarketItem : MarketItem {
-    public required long Id { get; init; }
-    public required UgcMarketListingStatus Status { get; init; }
-    public required long ListingEndTime { get; init; }
-    public required long PromotionEndTime { get; init; }
+public class
+    UgcMarketItem : MarketItem {
```
kOchirasu, line 10: nit: this should be on the same line

Maple2.Model/Game/Market/UgcMarketItem.cs
```diff
@@ -45,7 +49,7 @@ public class UgcMarketItem : MarketItem {
         writer.WriteLong(SellerCharacterId);
         writer.WriteUnicodeString();
         writer.WriteUnicodeString(SellerCharacterName);
-        writer.WriteUnicodeString(string.Join(",", Tags + ", " + Look.Name));
+        writer.WriteUnicodeString(string.Join(",", string.Join(",", Tags) + ", " + Look.Name));
```
kOchirasu, line 52: bit confused by the additional `string.Join` here, isn't the 2nd param just a single result already?

Maple2.Server.Core/Packets/UgcPacket.cs
```diff
@@ -42,4 +72,22 @@ private enum Command : byte {
...
+        var pWriter = Packet.Of(SendOp.Ugc);
+        pWriter.Write<Command>(Command.UpdateItem);
+        pWriter.WriteInt(objectId);
+        
+        pWriter.WriteLong(item.Uid);
+        pWriter.WriteInt(item.Id);
+        pWriter.WriteInt(item.Amount);
+        pWriter.WriteUnicodeString(item.Template!.Name);
+        pWriter.WriteByte(1);
+        pWriter.WriteLong(1); // price
```
kOchirasu, line 86: price is always 1?

Maple2.Server.Game/Manager/Config/ConfigManager.cs
```diff
@@ -138,6 +141,14 @@ public class ConfigManager {
         session.Send(AttributePointPacket.Allocation(statAttributes));
     }
 
+    public IList<long> GetFavoriteDesigners() {
+        return favoriteDesigners;
+    }
+
+    public void UpdateFavoriteDesigners(IList<long> designers) {
```
kOchirasu, line 148: `Add`+`Remove` seems more appropriate to me rather than overwriting the entire list on every modification. Then you can declare as `private readonly IList<long> favoriteDesigners`

Maple2.Server.Game/Manager/UgcMarketManager.cs
```diff
@@ -0,0 +1,160 @@
...
+
+    public void RefreshSales() {
+        using GameStorage.Request db = session.GameStorage.Context();
+        Save(db);
+        sales.Clear();
+        sales = db.GetMySoldUgcListings(session.AccountId);
+        session.Send(MeretMarketPacket.LoadSales(sales.Values));
+    }
+
+    public void Add(UgcMarketItem item) {
```
kOchirasu, line 54: `ListItem` might be a better name than `Add` since it's the counterpart of `UnlistItem`?

Maple2.Server.Game/Manager/UgcMarketManager.cs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    public void AddDesigner(long characterId) {
+        if (favoriteDesigners.ContainsKey(characterId)) {
+            LoadDesignerItems(characterId);
+            return;
+        }
+
+        using GameStorage.Request db = session.GameStorage.Context();
+        PlayerInfo? info = db.GetPlayerInfo(characterId);
```
kOchirasu, line 114: Avoid reading `PlayerInfo` directly from db, use `session.PlayerInfo.GetOrFetch(...)` so it's cached (and less stale)

Maple2.Server.Game/PacketHandlers/UgcHandler.cs
```diff
@@ -61,25 +67,97 @@ private enum Command : byte {
...
+                    break;
+                default:
+                    Logger.Error("Unhandled currency type {UgcMetadataCurrencyType}", ugcMetadata.CurrencyType);
+                    return;
+            }
+        }
+
+        Item? item = session.Item.CreateItem(itemId, ugcMetadata.ItemRarity);
+        if (item == null) {
+            return;
```
kOchirasu, line 136: This needs to be logged as fatal since the player already paid for this we shouldn't silently fail.

something similar to:
```cs
Log.Fatal("Failed to consume ingredient {ItemUid}", ingredient.Uid);
throw new InvalidOperationException($"Fatal: Consuming ingredient: {ingredient.Uid}");
```

Maple2.Server.Game/PacketHandlers/UgcHandler.cs
```diff
@@ -61,25 +67,97 @@ private enum Command : byte {
...
         }
 
-        // using WebStorage.Request request = WebStorage.Context();
-        // request.CreateUgc(info.CharacterId, "/path");
+        using WebStorage.Request request = WebStorage.Context();
+        UgcResource? resource = request.CreateUgc(ugcType, session.CharacterId);
+        if (resource == null) {
+            return;
```
kOchirasu, line 142: same as above, this needs to be logged

Maple2.Server.Game/Packets/MeretMarketPacket.cs
```diff
@@ -17,19 +18,150 @@ private enum Command : byte {
...
     }
 
+    public static ByteWriter LoadPersonalListings(ICollection<UgcMarketItem> items) {
+        var pWriter = Packet.Of(SendOp.MeretMarket);
+        pWriter.Write<Command>(Command.LoadPersonalListings);
+        pWriter.WriteLong();
+        pWriter.WriteInt(items.Count);
+
+        foreach (UgcMarketItem item in items) {
+            pWriter.WriteBool(true);
```
kOchirasu, line 42: Include a comment for what these bools are (if you know)

Maple2.Server.Game/Packets/MeretMarketPacket.cs
```diff
@@ -17,19 +18,150 @@ private enum Command : byte {
...
+        pWriter.WriteLong(id);
+        pWriter.WriteLong(id);
+
+        return pWriter;
+    }
+
+    public static ByteWriter RelistItem(UgcMarketItem item) {
+        var pWriter = Packet.Of(SendOp.MeretMarket);
+        pWriter.Write<Command>(Command.RelistItem);
+        pWriter.WriteBool(true);
```
kOchirasu, line 82: Same as above, try to document hardcoded values so they have a meaning.

Maple2.Server.Web/Controllers/WebController.cs
```diff
@@ -44,22 +44,25 @@ public class WebController : ControllerBase {
...
+        UgcResource? resource = null;
         if (ugcUid != 0) {
             using WebStorage.Request db = webStorage.Context();
-            UgcResource? resource = db.GetUgc(ugcUid);
+            resource = db.GetUgc(ugcUid);
             if (resource == null) {
                 return Results.NotFound($"{ugcUid} does not exist.");
             }
-            if (System.IO.File.Exists(resource.Path)) {
+            /*if (System.IO.File.Exists(resource.Path)) {
```
kOchirasu, line 54: is this intentionally commented?

Maple2.Server.World/Migrations/20230911002445_UgcMarket.cs
```diff
@@ -0,0 +1,107 @@
...
+                defaultValue: 0);
+
+            migrationBuilder.AddColumn<bool>(
+                name: "Giftable",
+                table: "premium-market-item",
+                type: "tinyint(1)",
+                nullable: false,
+                defaultValue: false);
+
+            migrationBuilder.AddColumn<string>(
```
kOchirasu, line 40: missing `defaultValue`, this is needed for all `AddColumn` to avoid breaking old rows.

Maple2.Model/Game/Market/UgcMarketItem.cs
Zintixx: The additional string.Join was due to needing to write each out item in the Tags array.

Maple2.Server.Web/Controllers/WebController.cs
Zintixx: Ah I was having issues with it due to Item and Item Icon colliding (both get uploaded one after another). I pushed an update that should address this.

Maple2.Database/Storage/Game/GameStorage.Market.cs
Zintixx: Fixed. I didn't do this for `GetUgcMarketPromotedItems` because it looks like it doesn't support Random.

kOchirasu: Try this:
image

PR #55, opened 2023-09-11
Skill Cooldowns

Zintixx:
- This is part of the #42 PR. Just breaking down that PR in chunks.


Maple2.Server.World/Migrations/20230911045230_SkillCooldown.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Maple2.Server.World.Migrations
+{
+    /// <inheritdoc />
+    public partial class SkillCooldown : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
```
kOchirasu, line 13: needs default value

PR #56, opened 2023-09-11
Gathering/Life Skills & More ConditionTypes

Zintixx:


Maple2.Server.Game/Manager/MasteryManager.cs
```diff
@@ -88,6 +94,81 @@ public class MasteryManager {
...
+            return;
+        }
+
+        if (recipeMetadata.RequiredMastery > this[recipeMetadata.Type]) {
+            session.Send(MasteryPacket.Error(MasteryError.s_mastery_error_lack_mastery));
+            return;
+        }
+
+        if (!gatheringCounts.TryGetValue(recipeMetadata.Id, out int currentCount)) {
+            gatheringCounts[recipeMetadata.Id] = currentCount;
```
kOchirasu, line 112: `currentCount` is always 0, set to 0 to be more clear.

```cs
gatheringCounts[recipeMetadata.Id] = 0;
```

Maple2.Server.Game/Manager/MasteryManager.cs
```diff
@@ -88,6 +94,81 @@ public class MasteryManager {
...
+        }
+
+        // TODO: Implement the bool on the last parameter of the formula. Is in someone else's home ?
+        float successRate = lua.CalcGatheringObjectSuccessRate(currentCount, recipeMetadata.HighRateLimitCount, recipeMetadata.NormalRateLimitCount);
+
+        int gatheringAmount = 0;
+        switch (recipeMetadata.Type) {
+            case MasteryType.Farming:
```
kOchirasu, line 120: Can we combine these case statements like Gathering+Mining? Also what's even the difference between harvest/farming?

Maple2.Model/Enum/Interact.cs
```diff
@@ -22,6 +22,7 @@ public enum InteractState : byte {
 
 // ReSharper disable InconsistentNaming, IdentifierTypo
 public enum InteractResult : byte {
+    none = 0,
```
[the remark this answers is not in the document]
Zintixx: right. 0 is used for telescopes as well. The enum has to be a byte so I wouldn't be able to do -1 (unless we cast on the packet).
I haven't tested how the client would act if the packet was sent as -1 but it is sent as 0 on sniffs.

Maple2.Server.Game/Manager/MasteryManager.cs
Zintixx: There's almost no difference. It's very weird but each have different condition types.

Maple2.Model/Enum/Interact.cs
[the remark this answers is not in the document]
kOchirasu: That's fine then. I kind of assumed this was just used for the code and not actually sent to clients. If it's being sent in a packet and expected to have a value of 0 to mean `none` then leave it as-is.

PR #58, opened 2023-09-13
Add base party implementation

skyrossm:
Working functionality:
- Inviting
- Joining
- Kicking
- Change Leader
- Request to join
- Offline/online

### DIFF
--- a/Maple2.Model/Game/Party/Party.cs
+++ b/Maple2.Model/Game/Party/Party.cs
@@ -8,8 +8,7 @@ using Maple2.Tools.Extensions;
 
 namespace Maple2.Model.Game;
 
-public class Party : IByteSerializable
-{
+public class Party : IByteSerializable {
     public int Capacity = 10;
 
     public required int Id { get; init; }
@@ -25,8 +24,7 @@ public class Party : IByteSerializable
     public readonly ConcurrentDictionary<long, PartyMember> Members;
 
     [SetsRequiredMembers]
-    public Party(int id, long leaderAccountId, long leaderCharacterId, string leaderName)
-    {
+    public Party(int id, long leaderAccountId, long leaderCharacterId, string leaderName) {
         Id = id;
         LeaderAccountId = leaderAccountId;
         LeaderCharacterId = leaderCharacterId;
@@ -39,16 +37,14 @@ public class Party : IByteSerializable
     [SetsRequiredMembers]
     public Party(int id, PartyMember leader) : this(id, leader.AccountId, leader.CharacterId, leader.Name) { }
 
-    public void WriteTo(IByteWriter writer)
-    {
+    public void WriteTo(IByteWriter writer) {
         writer.WriteBool(true);
         writer.WriteInt(Id);
         writer.WriteLong(LeaderCharacterId);
 
-        byte memberCount = (byte)Members.Count;
+        byte memberCount = (byte) Members.Count;
         writer.WriteByte(memberCount);
-        for (var i = 0; i < memberCount; i++)
-        {
+        for (var i = 0; i < memberCount; i++) {
             PartyMember member = Members.Values.ElementAt(i);
 
             writer.WriteBool(!member.Info.Online);
@@ -63,11 +59,9 @@ public class Party : IByteSerializable
         WriteMatchParty(writer);
     }
 
-    public void WriteMatchParty(IByteWriter writer)
-    {
+    public void WriteMatchParty(IByteWriter writer) {
         writer.WriteBool(IsMatching);
-        if (!IsMatching)
-        {
+        if (!IsMatching) {
             return;
         }
 
--- a/Maple2.Model/Game/Party/Party.cs
+++ b/Maple2.Model/Game/Party/Party.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using Maple2.Model.Metadata;
 using Maple2.PacketLib.Tools;
 using Maple2.Tools;
 using Maple2.Tools.Extensions;
@@ -9,7 +10,8 @@ using Maple2.Tools.Extensions;
 namespace Maple2.Model.Game;
 
 public class Party : IByteSerializable {
-    public int Capacity = 10;
+    private int capacity = Constant.PartyMaxCapacity;
+    public int Capacity { get => capacity; set { capacity = Math.Clamp(value, Constant.PartyMinCapacity, Constant.PartyMaxCapacity); } }
 
     public required int Id { get; init; }
     public required long LeaderAccountId;
@@ -44,9 +46,7 @@ public class Party : IByteSerializable {
 
         byte memberCount = (byte) Members.Count;
         writer.WriteByte(memberCount);
-        for (var i = 0; i < memberCount; i++) {
-            PartyMember member = Members.Values.ElementAt(i);
-
+        foreach (PartyMember member in Members.Values) {
             writer.WriteBool(!member.Info.Online);
             writer.WriteClass(member);
             member.WriteDungeonEligibility(writer);
--- a/Maple2.Model/Game/Party/Party.cs
+++ b/Maple2.Model/Game/Party/Party.cs
@@ -11,7 +11,12 @@ namespace Maple2.Model.Game;
 
 public class Party : IByteSerializable {
     private int capacity = Constant.PartyMaxCapacity;
-    public int Capacity { get => capacity; set { capacity = Math.Clamp(value, Constant.PartyMinCapacity, Constant.PartyMaxCapacity); } }
+    public int Capacity {
+        get => capacity;
+        set {
+            capacity = Math.Clamp(value, Constant.PartyMinCapacity, Constant.PartyMaxCapacity);
+        }
+    }
 
     public required int Id { get; init; }
     public required long LeaderAccountId;
@@ -40,7 +45,7 @@ public class Party : IByteSerializable {
     public Party(int id, PartyMember leader) : this(id, leader.AccountId, leader.CharacterId, leader.Name) { }
 
     public void WriteTo(IByteWriter writer) {
-        writer.WriteBool(true);
+        writer.WriteBool(true); // joining from offline?
         writer.WriteInt(Id);
         writer.WriteLong(LeaderCharacterId);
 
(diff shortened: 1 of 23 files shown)


Maple2.Server.Game/PacketHandlers/PartyHandler.cs
```diff
@@ -0,0 +1,249 @@
...
+                Reply = (int) response,
+            },
+        });
+        var error = (PartyError) partyResponse.Error;
+        if (error != PartyError.none) {
+            session.Send(PartyPacket.Error(error));
+            return;
+        }
+
+        if (response == PartyInvite.Response.Accept) {
```
Zintixx, line 151: We'll also need to check if the party still exists.

Maple2.Server.World/Containers/PartyManager.cs
```diff
@@ -0,0 +1,216 @@
...
+
+            try {
+                client.Party(request);
+            } catch { }
+        }
+    }
+
+    public void FindNewLeader(long characterId) {
+        PartyMember? newLeader = Party.Members.Values.FirstOrDefault(m => m.CharacterId != characterId);
+        if (newLeader == null) {
```
Zintixx, line 47: Maybe disband here? I can't think of an instance where the party should still exist if this happens.

Maple2.Model/Game/Party/PartyMember.cs
```diff
@@ -0,0 +1,90 @@
...
+    public long CharacterId => Info.CharacterId;
+    public string Name => Info.Name;
+    public byte ReadyState = 0;
+
+    public CancellationTokenSource? TokenSource;
+
+    public void WriteTo(IByteWriter writer) {
+        writer.WriteLong(Info.AccountId);
+        writer.WriteLong(Info.CharacterId);
+        writer.WriteUnicodeString(Info.Name);
```
Zintixx, line 26: shouldn't we just be using Character? I believe it's the same struct.

Maple2.Server.Game/PacketHandlers/PartyHandler.cs
Zintixx: take this scenario:
Player 1 invites Player 2 to their party.
Player 1 decides to disband (for whatever reason, logged out, their decision, joined a different party)
Player 2 still has the prompt to join Player 1's party


If Player 1 accepts, it should return that the party no longer exists.

Maple2.Model/Game/Party/Party.cs
```diff
@@ -0,0 +1,81 @@
...
+    public Party(int id, PartyMember leader) : this(id, leader.AccountId, leader.CharacterId, leader.Name) { }
+
+    public void WriteTo(IByteWriter writer) {
+        writer.WriteBool(true);
+        writer.WriteInt(Id);
+        writer.WriteLong(LeaderCharacterId);
+
+        byte memberCount = (byte) Members.Count;
+        writer.WriteByte(memberCount);
+        for (var i = 0; i < memberCount; i++) {
```
kOchirasu, line 47: generally prefer `foreach` loops if you can. It's less bug prone.

Maple2.Model/Game/Party/Party.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        writer.WriteBool(false);
+        writer.WriteInt(DungeonId);
+        writer.WriteBool(false);
+        writer.WriteByte();
+        WriteMatchParty(writer);
+    }
+
+    public void WriteMatchParty(IByteWriter writer) {
```
kOchirasu, line 62: private?

Maple2.Model/Game/Party/PartyMember.cs
kOchirasu: +1, use PlayerInfo if you can. It has less data so it will be less stale, and there's infrastructure to keep it updated.

Maple2.Model/Game/Party/PartyMember.cs
```diff
@@ -0,0 +1,90 @@
...
+        writer.WriteInt((int) Info.Job.Code());
+        writer.Write(Info.Job);
+        writer.WriteInt((int) Info.CurrentHp);
+        writer.WriteInt((int) Info.TotalHp);
+        writer.WriteShort();
+        writer.WriteLong();
+        writer.WriteLong();
+        writer.WriteLong();
+        writer.WriteInt();
+        writer.Write(new Vector3());
```
kOchirasu, line 45: Here and elsewhere, for structs you can just use `default`. Also for structs always include type in `Write` call to prevent bugs where someone accidentally changes the type. (Basically there should be no instances of `writer.Write(...)`)

Example:
```cs
writer.Write<Vector3>(default);
```

Maple2.Model/Game/Party/PartyMember.cs
```diff
@@ -0,0 +1,90 @@
...
+    public long JoinTime;
+    public long LoginTime;
+    public long AccountId => Info.AccountId;
+    public long CharacterId => Info.CharacterId;
+    public string Name => Info.Name;
+    public byte ReadyState = 0;
+
+    public CancellationTokenSource? TokenSource;
+
+    public void WriteTo(IByteWriter writer) {
```
kOchirasu, line 23: Kinda surprised the whole character struct is here, but I guess it still works to just write 0 values?

Maple2.Server.Game/PacketHandlers/PartyHandler.cs
```diff
@@ -0,0 +1,249 @@
...
+
+            var error = (PartyError) response.Error;
+            if (error != PartyError.none) {
+                session.Send(PartyPacket.Error(error));
+                return;
+            }
+
+            session.Party.SetParty(response.Party);
+        }
+
```
kOchirasu, line 104: we should validate that the player sending the invite is the leader of the party.

Maple2.Server.Game/PacketHandlers/PartyHandler.cs
kOchirasu: Player 1 would still have the prompt, but when Player 1 accepts, it will make a request to `WorldServer` to join the party. At this point, `WorldServer` knows the party is disbanded, so the request to join the party *should* fail and Player 1 will get an error instead.

Maple2.Server.Game/PacketHandlers/PartyHandler.cs
```diff
@@ -0,0 +1,249 @@
...
+            Kick = new PartyRequest.Types.Kick {
+                PartyId = session.Party.Id,
+                ReceiverId = targetCharacterId,
+            },
+        });
+
+        var error = (PartyError) response.Error;
+        if (error != PartyError.none) {
+            return;
+        }
```
kOchirasu, line 183: `RemoveMember()` should be called after this

Maple2.Server.World/Containers/PartyManager.cs
```diff
@@ -0,0 +1,216 @@
...
+        return PartyError.none;
+    }
+
+    public PartyError Leave(long characterId) {
+        if (!Party.Members.TryGetValue(characterId, out PartyMember? member)) {
+            return PartyError.s_party_err_not_found;
+        }
+
+        if (Party.Members.Count <= 2) {
+            Broadcast(new PartyRequest {
```
kOchirasu, line 154: might make sense to just move this `Broadcast(...)` into `Dispose()`. You're currently missing this Broadcast when the party is disbanded naturally.

Maple2.Model/Game/Party/Party.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using Maple2.PacketLib.Tools;
+using Maple2.Tools;
+using Maple2.Tools.Extensions;
+
+namespace Maple2.Model.Game;
+
+public class Party : IByteSerializable {
+    public int Capacity = 10;
```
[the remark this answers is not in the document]
kOchirasu: I see, the `MaxCapacity` should be a constant. When setting capacity, we could do something like `Min(capacity, MaxCapacity)`.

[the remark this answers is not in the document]
kOchirasu: yea you're right, should be 1 or whatever the actual min is. Maybe that should be a constant as well.

Maple2.Model/Game/Party/Party.cs
kOchirasu: makes sense

Maple2.Server.Game/PacketHandlers/PartyHandler.cs
kOchirasu: Oh, I see how it works now.

Maple2.Model/Game/Party/Party.cs
```diff
@@ -0,0 +1,81 @@
...
+    public void WriteTo(IByteWriter writer) {
+        writer.WriteBool(true);
+        writer.WriteInt(Id);
+        writer.WriteLong(LeaderCharacterId);
+
+        byte memberCount = (byte) Members.Count;
+        writer.WriteByte(memberCount);
+        foreach (PartyMember member in Members.Values) {
+            writer.WriteBool(!member.Info.Online);
+            writer.WriteClass(member);
```
kOchirasu, line 51: nit: `writer.WriteClass<PartyMember>(member);`

Maple2.Model/Game/Party/PartyMember.cs
```diff
@@ -0,0 +1,90 @@
...
+        writer.WriteLong(Info.AccountId);
+        writer.WriteLong(Info.CharacterId);
+        writer.WriteUnicodeString(Info.Name);
+        writer.Write(Info.Gender);
```
kOchirasu, line 27: nit: `writer.Write<Gender>(Info.Gender);`

Maple2.Model/Game/Party/PartyMember.cs
```diff
@@ -0,0 +1,90 @@
...
+        writer.WriteByte(1);
+        writer.WriteLong();
+        writer.WriteInt();
+        writer.WriteInt(Info.MapId);
+        writer.WriteInt(Info.MapId);
+        writer.WriteInt(Info.PlotMapId);
+        writer.WriteShort(Info.Level);
+        writer.WriteShort(Info.Channel);
+        writer.WriteInt((int) Info.Job.Code());
+        writer.Write(Info.Job);
```
kOchirasu, line 37: nit: `writer.Write<Job>(Info.Job);`

Maple2.Model/Game/Party/PartyMember.cs
```diff
@@ -0,0 +1,90 @@
...
+        writer.Write<SkinColor>(default);
+        writer.WriteLong();
+        writer.Write<AchievementInfo>(default);
+        writer.WriteLong();
+        writer.WriteUnicodeString();
+        writer.WriteUnicodeString(Info.Motto);
+        writer.WriteUnicodeString(Info.Picture);
+        writer.WriteByte();
+        writer.WriteByte();
+        writer.WriteClass(new Mastery());
```
kOchirasu, line 56: nit: `writer.WriteClass<Mastery>(new Mastery());`

Maple2.Model/Game/Party/Party.cs
```diff
@@ -0,0 +1,81 @@
...
+using Maple2.Model.Metadata;
+using Maple2.PacketLib.Tools;
+using Maple2.Tools;
+using Maple2.Tools.Extensions;
+
+namespace Maple2.Model.Game;
+
+public class Party : IByteSerializable {
+    private int capacity = Constant.PartyMaxCapacity;
+    public int Capacity { get => capacity; set { capacity = Math.Clamp(value, Constant.PartyMinCapacity, Constant.PartyMaxCapacity); } }
```
kOchirasu, line 14: nit: multiline so it's easier to read

```cs
public int Capacity {
    get => capacity;
    set {
        capacity = Math.Clamp(value, Constant.PartyMinCapacity, Constant.PartyMaxCapacity);
    }
}
```

Maple2.Model/Game/Party/Party.cs
```diff
@@ -0,0 +1,81 @@
...
+            writer.WriteBool(!member.Info.Online);
+            writer.WriteClass(member);
+            member.WriteDungeonEligibility(writer);
+        }
+
+        writer.WriteBool(false);
+        writer.WriteInt(DungeonId);
+        writer.WriteBool(false);
+        writer.WriteByte();
+        WriteMatchParty(writer);
```
kOchirasu, line 59: This should be outside of the `Party` struct, by including it inside, it means anytime you write `Party` it must be followed by `MatchParty`. (the byte above should be moved as well)

see: https://github.com/kOchirasu/MapleShark2-Scripts/blob/932829f91d11c1c96ab3531b08268351c58382a4/Inbound/0x0054.py#L188C1-L190C37

PR #59, opened 2023-09-21
Shops pt 2

Zintixx:
- Implements player specific shop data in order to handle stock counts
- Implements buy/sell/buy back of shops
- Refreshing shop/instant restocking 
- Implements buy period of items in shops. No current item in the game uses this function though.
- Contains about 90% of all shops in the game. the 10% is misc event shops and some deprecated shops (such as apartment complex shops, furnishing shops, etc)


Maple2.Server.Game/GameServer.cs
```diff
@@ -87,6 +93,54 @@ public GameServer(FieldManager.Factory fieldFactory, PacketRouter<GameSession> r
...
+            if (shop != null) {
+                shopCache[shopId] = shop;
+            }
+            if (shop?.RestockTime == 0) { // everything else would be player-based shops that would get refreshed
+                if (!shopItemCache.TryGetValue(shopId, out Dictionary<int, ShopItem>? shopItems)) {
+                    shopItems = db.GetShopItems(shopId).ToDictionary(item => item.Id);
+                    shopItemCache[shopId] = shopItems;
+                }
+                foreach ((int shopItemId, ShopItem shopItem) in shopItems) {
+                    Item? item = session.Item.CreateItem(shopItem.ItemId, shopItem.Rarity, shopItem.Quantity);
```
kOchirasu, line 109: this causes the item to be created in db, is that necessary for shop items before they are purchased? (do they need UID) it would flood the database with useless entries.

Maple2.Server.Game/GameServer.cs
```diff
@@ -87,6 +93,54 @@ public GameServer(FieldManager.Factory fieldFactory, PacketRouter<GameSession> r
...
+                    shop.Items[shopItem.Id] = shopItem;
+                }
+            } else {
+                return shop;
+            }
+        }
+
+        if (shop.Items.Count == 0 && shopItemCache.TryGetValue(shop.Id, out Dictionary<int, ShopItem>? items)) {
+            foreach ((int shopItemId, ShopItem shopItem) in items) {
+                Item? item = session.Item.CreateItem(shopItem.ItemId, shopItem.Rarity, shopItem.Quantity);
```
kOchirasu, line 123: same as above, i would prefer to not create these items in db until after purchase.

Maple2.Server.Game/PacketHandlers/ShopHandler.cs
```diff
@@ -21,20 +21,57 @@ private enum Command : byte {
...
+
+    private void HandleSell(GameSession session, IByteReader packet) {
+        long itemUid = packet.ReadLong();
+        int quantity = packet.ReadInt();
+
+        session.Shop.Sell(itemUid, quantity);
+    }
+
+    private void HandleInstantRestock(GameSession session, IByteReader packet) {
+        packet.ReadInt(); // cost
```
kOchirasu, line 69: missing code to charge the user?

Maple2.Server.World/Migrations/20230921034534_ShopsPart2.cs
```diff
@@ -0,0 +1,238 @@
...
+
+#nullable disable
+
+namespace Maple2.Server.World.Migrations
+{
+    /// <inheritdoc />
+    public partial class ShopsPart2 : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
```
kOchirasu, line 12: include defaultValue where needed (AddColumn)

Maple2.Server.Game/GameServer.cs
Zintixx: I don't think it does get created in db. CreateItem doesn't add it to db.

Maple2.Server.Game/PacketHandlers/ShopHandler.cs
Zintixx: Nah. It's not using the packet read here to ensure the current price is correct using ShopManager.InstantRestock

Maple2.Server.Game/GameServer.cs
kOchirasu: You're right, i think i looked at the wrong function.

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+using Serilog;
+
+namespace Maple2.Server.Game.Manager;
+
+public sealed class ShopManager {
+    #region EntryId
+    private int idCounter;
+
+    /// <summary>
+    /// Generates an EntryId unique to this specific actor instance.
```
kOchirasu, line 26: typo: "manager instance"

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+
+    private readonly ILogger logger = Log.Logger.ForContext<ShopManager>();
+
+    public ShopManager(GameSession session) {
+        this.session = session;
+        instancedShops = new Dictionary<int, Shop>();
+        buyBackItems = new Dictionary<int, BuyBackItem>();
+        using GameStorage.Request db = session.GameStorage.Context();
+
+        accountShopData = new Dictionary<int, CharacterShopData>();
```
kOchirasu, line 50: You don't need to copy the items from one dictionary to another. If you really want a copy, you can use the copy constructor: `new Dictionary(other)`, but why can't you just use the returned value directly?

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+            }
+            return CreateInstancedShop(serverShop, CreateShopData(serverShop), restockTime);
+
+        }
+        if (data.RestockTime < DateTime.Now.ToEpochSeconds()) {
+            return CreateInstancedShop(serverShop, CreateShopData(serverShop), GetRestockTime(serverShop.RestockData!.Interval));
+        }
+
+        // Assemble shop
+        instancedShop = serverShop.Clone()!;
```
kOchirasu, line 151: here and elsewhere, why do you need `!`? how can this be null.

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+        instancedShop.RestockData!.RestockCount = shopData.RestockCount;
+        CreateShopItemData(instancedShop);
+        instancedShops[instancedShop.Id] = instancedShop;
+        return instancedShop;
+    }
+
+    private SortedDictionary<int, ShopItem> GetShopItems(Shop shop) {
+        var items = new SortedDictionary<int, ShopItem>();
+        using GameStorage.Request db = session.GameStorage.Context();
+        if (shop.RestockData?.Interval == ShopRestockInterval.Minute) {
```
kOchirasu, line 198: seems like you can simplify this:
```cs
IEnumerable<ShopItem> itemList;
if (shop.RestockData?.Interval == ShopRestockInterval.Minute) {
    itemList = session.FindShopItems(shop.Id).OrderBy(_ => Random.Shared.Next()).Take(12);
} else {
    itemList = shop.Items.Values;
}

foreach (ShopItem shopItem in itemList) {
    Item? item = session.Item.CreateItem(shopItem.ItemId, shopItem.Rarity, shopItem.Quantity);
    if (item == null) {
        continue;
    }
    shopItem.Item = item;
    items.Add(shopItem.Id, shopItem.Clone());
}
```

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+            foreach ((int shopItemId, CharacterShopItemData itemData) in shopItemDatas) {
+                db.DeleteCharacterShopItemData(ownerId, shopItemId);
+            }
+        }
+
+        // Create new shop data
+        instancedShop.Items = GetShopItems(shop);
+        instancedShop.RestockTime = restockTime;
+        instancedShop.RestockData!.RestockCount = shopData.RestockCount;
+        CreateShopItemData(instancedShop);
```
kOchirasu, line 190: return value of this function is unused?

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+            if (shop.RestockData.Interval != ShopRestockInterval.Minute) {
+                data = db.CreateCharacterShopItemData(ownerId, data);
+                if (data == null) {
+                    continue;
+                }
+            }
+            itemData[id] = data;
+        }
+
+        if (shop.RestockData!.PersistantInventory) {
```
kOchirasu, line 271: `shop.RestockData?.PersistantInventory == true` seems safer, you should use `!` more as last resort

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+            accountShopItemData[shop.Id] = itemData;
+            return accountShopItemData[shop.Id];
+        }
+        characterShopItemData[shop.Id] = itemData;
+        return characterShopItemData[shop.Id];
+    }
+
+    public void InstantRestock() {
+        if (activeShop?.RestockData == null || activeShop.RestockData.DisableInstantRestock) {
```
kOchirasu, line 280: `activeShop?.RestockData?.DisableInstantRestock != false` should be equivalent

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+        ShopCurrencyType currencyType = activeShop.RestockData.CurrencyType;
+        if (activeShop.RestockData.EnableCostMultiplier) {
+            (ShopCurrencyType CurrencyType, int Cost) multiplierCost = Core.Formulas.Shop.ExcessRestockCost(activeShop.RestockData);
+            cost = multiplierCost.Cost;
+            currencyType = multiplierCost.CurrencyType;
+        }
+        if (!Pay(new ShopCost {
+                Amount = cost,
+                Type = currencyType,
+                ItemId = 0,
```
kOchirasu, line 294: should we set some defaults so you don't have to specify 0 here?

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+                SaleAmount = 0,
+            }, activeShop.RestockData.Cost)) {
+            return;
+        }
+
+        Shop? shop = session.FindShop(activeShop.Id);
```
kOchirasu, line 300: we should probably load this shop first so we don't risk failure after user has already paid.

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+
+            if (buyData.Days.Count > 0) {
+                if (!buyData.Days.Contains(ToShopBuyDay(DateTime.Now.DayOfWeek))) {
+                    session.Send(ShopPacket.Error(ShopError.s_err_invalid_item_cannot_buy_by_period));
+                    return;
+                }
+            }
+        }
+
+        if (shopItem.StockCount > 0 && shopItem.StockPurchased + quantity > shopItem.StockCount) {
```
kOchirasu, line 407: I'm worried adding `shopItem.StockPurchased + quantity` given `quantity` is user input, you can cause overflow. I think a better check is something like `quantity > shopItem.StockCount - shopItem.StockPurchased`

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+            return;
+        }
+
+        session.Currency.Meso += sellPrice;
+
+        if (buyBackItems.Count >= Constant.MaxBuyBackItems) {
+            RemoveBuyBackItem();
+        }
+
+        if (!session.Item.Inventory.Remove(item.Uid, out item, quantity)) {
```
kOchirasu, line 497: seems better to remove this before adding meso to avoid exploits

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+
+        if (!Pay(new ShopCost {
+                Type = ShopCurrencyType.Meso,
+                Amount = (int) buyBackItem.Price,
+            }, (int) buyBackItem.Price)) {
+            return;
+        }
+
+        session.Item.Inventory.Add(buyBackItem.Item, true);
+        buyBackItems.Remove(id);
```
kOchirasu, line 467: I would generally prefer removing before adding to avoid exploits (and the remove should succeed)

Maple2.Server.Game/Manager/ShopManager.cs
```diff
@@ -0,0 +1,609 @@
...
+        }
+        return true;
+    }
+
+    private void RemoveBuyBackItem() {
+        while (buyBackItems.Count >= Constant.MaxBuyBackItems) {
+            BuyBackItem? item = buyBackItems.Values.MinBy(entry => entry.AddedTime);
+            if (item == null) {
+                logger.Error("Failed to remove buyback item");
+                return;
```
kOchirasu, line 577: return false here

Maple2.Server.Game/Manager/ShopManager.cs
Zintixx: This function is used in two instances: Creating an instanced shop and getting an instanced shop. We're only using this function to create the shop item data entries and not assigning upon creation because it would be an exact identical copy of the current shop.

PR #61, opened 2023-09-26
Pocket Realm Portals & Player Hosted Mini Game Portals

Zintixx:
- Implements pocket realm portals. My estimation of these portals is they spawn within 5 minute intervals anywhere from 5 minutes to 8 hours.
- Implements player hosted mini game portals.
- Also implements a quick warping command by using coordinates.
- I'm having to hardcode the models for the portals because I don't know where these models are located.

### DIFF
--- a/Maple2.Model/Enum/Portal.cs
+++ b/Maple2.Model/Enum/Portal.cs
@@ -1,0 +1,16 @@
+﻿namespace Maple2.Model.Enum;
+
+public enum PortalType : byte {
+    Field = 0,
+    DungeonReturnToLobby = 1,
+    Event = 3, // Pocket Realms & Player Hosted Mini games
+    Boss = 8,
+    DungeonEnter = 9,
+    Home = 11,
+    LeaveDungeon = 13,
+}
+
+public enum PortalActionType {
+    Interact = 0,
+    Touch = 1,
+}
--- a/Maple2.Model/Enum/Portal.cs
+++ b/Maple2.Model/Enum/Portal.cs
@@ -4,9 +4,10 @@ public enum PortalType : byte {
     Field = 0,
     DungeonReturnToLobby = 1,
     Event = 3, // Pocket Realms & Player Hosted Mini games
+    FieldToHome = 5,
     Boss = 8,
     DungeonEnter = 9,
-    Home = 11,
+    InHome = 11,
     LeaveDungeon = 13,
 }
 
--- a/Maple2.Server.Game/Manager/Field/FieldManager.State.cs
+++ b/Maple2.Server.Game/Manager/Field/FieldManager.State.cs
@@ -71,7 +71,7 @@ public partial class FieldManager {
         if (Metadata.Property.RevivalReturnId != 0) {
             player.Character.ReviveMapId = Metadata.Property.RevivalReturnId;
         }
-        if (Metadata.Property.EnterReturnId != 0) {
+        if (Metadata.Property.EnterReturnId != 0 && Metadata.Limit.Capacity > 0) {
             player.Character.ReturnMapId = Metadata.Property.EnterReturnId;
         }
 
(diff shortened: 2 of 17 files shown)


Maple2.Server.Game/Manager/Field/FieldManager.State.cs
```diff
@@ -365,6 +365,24 @@ public partial class FieldManager {
         Broadcast(FieldPropertyPacket.Background(background));
     }
 
+    private void SetBonusMapPortal(IList<MapMetadata> bonusMaps, Ms2RegionSpawn spawn) {
+        // Spawn a hat within a random range of 5 min to 8 hours
+        int delay = Random.Shared.Next(1, 97) * (int) TimeSpan.FromMinutes(5).TotalMilliseconds;
+        var portal = new Portal(NextLocalId(), bonusMaps[Random.Shared.Next(bonusMaps.Count)].Id,  -1, PortalType.Event, PortalActionType.Interact, spawn.Position, spawn.Rotation,
+            new Vector3(200, 200, 250), 0, true, false, true);
+        FieldPortal fieldPortal = SpawnPortal(portal);
+        fieldPortal.Model = Metadata.Property.Continent switch {
```
kOchirasu, line 374: you think there's an xblock we should be parsing for this data?

Maple2.Server.Game/PacketHandlers/ItemUseHandler.cs
```diff
@@ -421,4 +424,23 @@ public class ItemUseHandler : PacketHandler<GameSession> {
...
+        string password = packet.ReadUnicodeString();
+        Dictionary<string, string> parameters = XmlParseUtil.GetParameters(item.Metadata.Function?.Parameters);
+
+        // TODO: Handle maxCount and weddingParty
+        if (!parameters.ContainsKey("fieldID") || !int.TryParse(parameters["fieldID"], out int fieldId) ||
+            !parameters.ContainsKey("portalDurationTick") || !int.TryParse(parameters["portalDurationTick"], out int portalDurationTick)) {
+            return;
+        }
+
+        session.Item.Inventory.Consume(item.Uid, 1);
```
kOchirasu, line 441: handle `false` return here (failure)

Maple2.Server.Game/Packets/PortalPacket.cs
```diff
@@ -26,18 +27,18 @@ private enum Command : byte {
...
         pWriter.WriteInt(fieldPortal.ObjectId);
-        pWriter.WriteInt();
+        pWriter.Write<PortalActionType>(portal.ActionType);
         pWriter.WriteBool(fieldPortal.MinimapVisible);
-        pWriter.WriteLong();
-        pWriter.WriteByte(portal.Type);
-        pWriter.WriteInt(); // StartTick
+        pWriter.WriteLong(fieldPortal.HomeId);
+        pWriter.Write<PortalType>(portal.Type);
+        pWriter.WriteInt(fieldPortal.EndTick);
```
kOchirasu, line 37: are you sure? previous comment said `StartTick` and the int below is `EndTick`

Maple2.Server.Game/Util/XmlParseUtil.cs
```diff
@@ -10,6 +10,10 @@ public static class XmlParseUtil {
             return parameters;
         }
 
+        if (xmlString[1] == ' ') {
```
kOchirasu, line 13: you could use `Trim()` which is more robust

```cs
xmlString = xmlString.Trim()
```

Maple2.Server.Game/Manager/Field/FieldManager.State.cs
Zintixx: I looked at the xblocks and no dice. I think these are within Model.m2d.

Maple2.Server.Game/Packets/PortalPacket.cs
Zintixx: Yep, pretty positive about this. It wouldn't show the duration of the portal until I swapped it to EndTick

Maple2.Server.Game/Util/XmlParseUtil.cs
Zintixx: I don't think Trim here would work.

The problem isn't the first character, it's the 2nd. This is how the string is formatted in the item XML:
```< v fieldID="81000003" portalDurationTick="180000" maxCount="20" />```

That space between '<' and 'v' breaks loading it as an XML.

File: Maple2.Server.Game/Manager/Field/FieldManager.State.cs (hunk 1 above)
Comment (kOchirasu) [reply Maple2.Server.Game/Manager/Field/FieldManager.State.cs:374]: yea i guess xblock are only static models, and since these are server-sided only it doesnt show up. Nexon probably has something else that includes this data.

File: Maple2.Model/Enum/Portal.cs [hunk 5]
```diff
@@ -0,0 +1,16 @@
+﻿namespace Maple2.Model.Enum;
+
+public enum PortalType : byte {
```
kOchirasu, line 3: some notes:
Type 4 and 11 are the only ones that set `ActionType`
image

type `{4}` is `Portal_Type_A`
type `{3, 5, 6, 7, 10, 11, 16}` are `MS2RoomEnterPortal`

Zintixx: That makes sense for 11 (don't know what 4 is). Considering Portal is a record, how would you recommend adjusting this to set the action type?

kOchirasu: isn't action type already set from xml?

kOchirasu: if you need to mutate, it's the same as readonly structs and you can use `with` syntax to copy with mutation.

```cs
Portal a;
Portal b = a with {ActionType = VALUE};
```

Zintixx: Yes and no. Field portals set by the xblocks are already correctly set.
Portals from the hidden realms/bonus maps spawn objects are not found in the xblocks, hence why they're being set manually on FieldManager.SpawnEventPortal

But even then, those portals are Interact only (because they're portal type 3)

PR #62, opened 2023-09-28
Map Item Spawns, Exp Changes, Item Loot Changes

Zintixx:
- Ingests event spawn point items from xblock for triggered spawned items
- Changed packet structure a bit on exp up.
- Moved the IsCurrency switch from item pick up to Inventory.Add in order to handle meret from item boxes
- Bind items of newly created characters
- Item drops from Interacting with Objects


Maple2.Model/Enum/ExpMessageCode.cs
```diff
@@ -4,29 +4,74 @@
...
-    s_msg_take_taxi_exp = 3000,
-    [Description("You got {1} experience for discovering {0}.")]
-    s_msg_take_map_exp = 3001, // and 3002
-    [Description("You got {1} experience for discovering a new area, {0}.")]
-    s_msg_take_telescope_exp = 3003,
-    [Description("You got {0} bonus experience.")]
-    s_msg_take_assist_bonus_exp = 60002,
-    [Description("You got an experience bonus for helping defeat a powerful foe. (EXP +{0})")]
-    s_msg_take_assist_bonus_exp_system = 60002,
+public enum ExpMessageCode : ushort {
```
kOchirasu, line 7: the previous enum seems more accurate for what this is being used for. You can have a separate function mapping these new values to the enum

PR #63, opened 2023-09-29
Buffs: part 2

Zintixx:
This is just #42  but the buff additions only.


Maple2.Server.Game/Manager/Config/BuffManager.cs
```diff
@@ -143,6 +168,128 @@ public class BuffManager : IUpdatable {
...
+            if (Invokes.TryGetValue(buff.Metadata.InvokeEffect.Types[i], out IDictionary<int, InvokeRecord>? nestedInvokeDic)) {
+                Invokes.RemoveAll(buff.Id);
+
+                if (!nestedInvokeDic.TryAdd(buff.Id, record)) {
+                    logger.Error("Could not add invoke record from {Id} to {Object}", buff.Id, actor.ObjectId);
+                }
+                continue;
+            }
+
+            if (!Invokes.TryAdd(buff.Metadata.InvokeEffect.Types[i], new ConcurrentDictionary<int, InvokeRecord> {
```
kOchirasu, line 193: Curious if `TryAdd` is expected to fail often here. Reason being we are instantiating a new dictionary only for it to fail. If it's expected to fail often, it would be better to defer the creation of the dictionary.

Maple2.Server.Game/Manager/Config/BuffManager.cs
```diff
@@ -143,6 +168,128 @@ public class BuffManager : IUpdatable {
...
+        if (Compulsions.TryGetValue(eventType, out IDictionary<int, AdditionalEffectMetadataStatus.CompulsionEvent>? nestedCompulsionDic)) {
+            Compulsions.RemoveAll(buff.Id);
+
+            if (!nestedCompulsionDic.TryAdd(buff.Id, buff.Metadata.Status.Compulsion)) {
+                logger.Error("Could not add compulsion event from {Id} to {Object}", buff.Id, actor.ObjectId);
+            }
+            return;
+        }
+
+        if (!Compulsions.TryAdd(eventType, new ConcurrentDictionary<int, AdditionalEffectMetadataStatus.CompulsionEvent> {
```
kOchirasu, line 216: same as above

Maple2.Server.Game/Manager/Config/BuffManager.cs
```diff
@@ -204,6 +351,20 @@ public class BuffManager : IUpdatable {
...
         }
+
+        foreach ((BasicAttribute attribute, float value) in buff.Metadata.Status.Resistances) {
+            Resistances[attribute] = Math.Min(0, Resistances[attribute] - value);
+        }
+
+        Invokes.RemoveAll(id);
+        Compulsions.RemoveAll(id);
+
+        if (buff.Metadata.Status.Values.Any() || buff.Metadata.Status.Rates.Any() || buff.Metadata.Status.SpecialValues.Any() || buff.Metadata.Status.SpecialRates.Any()) {
```
kOchirasu, line 362: We should use `.Count > 0` here. It's a lot faster.

```cs
if (buff.Metadata.Status.Values.Count > 0 || buff.Metadata.Status.Rates.Count > 0 || buff.Metadata.Status.SpecialValues.Count > 0 || buff.Metadata.Status.SpecialRates.Count > 0) {
```

https://www.tabsoverspaces.com/233649-comparing-speed-of-count-0-and-any

Maple2.Server.Game/Model/Field/Buff.cs
```diff
@@ -289,7 +290,7 @@ public class Buff : IUpdatable, IByteSerializable {
     }
 
     // Unknown, AdditionalEffect2
```
kOchirasu, line 292: you can remove this comment now since we know what this is.

Maple2.Server.Game/Trigger/TriggerContext.Npc.cs
```diff
@@ -112,7 +112,7 @@ public partial class TriggerContext {
     public void SetConversation(byte type, int spawnId, string script, int delay, byte arg5, Align align) {
         ErrorLog("[SetConversation] type:{Type}, spawnId:{SpawnId}, script:{Script}, delay:{Delay}, arg5:{Arg5}, align:{Align}",
             type, spawnId, script, delay, arg5, align);
-        Broadcast(CinematicPacket.Talk(spawnId, spawnId.ToString(), script, delay * 1000, align));
+        CinematicPacket.Talk(spawnId, spawnId.ToString(), script, delay * 1000, align);
```
kOchirasu, line 115: This packet isn't sent anymore?

Maple2.Server.Game/Manager/Config/BuffManager.cs
Zintixx: No it rarely should ever fail. I pushed a change to just use Add instead

PR #65, opened 2023-10-24
Group Chat

Zintixx:
There doesn't look to be a lot of errors to use, so I've defaulted in using `s_err_groupchat_null_target_user`


Maple2.Model/Error/GroupChatError.cs
```diff
@@ -0,0 +1,23 @@
+// ReSharper disable InconsistentNaming, IdentifierTypo
+
+using System.ComponentModel;
+
+namespace Maple2.Model.Error;
+
+public enum GroupChatError {
```
kOchirasu, line 7: `s_change_charname_err_bad_words = 13`

Maple2.Model/Error/GroupChatError.cs
```diff
@@ -0,0 +1,23 @@
...
+    none = 0,
+    [Description("You cannot invite that player to the group chat.")]
+    s_err_groupchat_null_target_user = 3,
+    [Description("You cannot invite that player to the group chat.")]
+    s_err_groupchat_add_member_target = 8,
+    [Description("{0} is already participating in 3 group chats.")]
+    s_err_groupchat_maxgroup = 10,
+
+    // Unknown values
```
kOchirasu, line 16: None of these are in the packet, they're probably client-side only.

Maple2.Model/Game/Party/PartyMember.cs
```diff
@@ -22,56 +22,7 @@ public class PartyMember : IByteSerializable, IDisposable {
...
-        writer.WriteInt();
-        writer.WriteInt();
-        writer.WriteLong(DateTimeOffset.UtcNow.ToUnixTimeSeconds());
-        writer.WriteInt();
-        writer.WriteLong();
-        writer.WriteInt();
-        writer.WriteInt();
-        writer.WriteShort();
-        writer.WriteLong();
+        writer.WriteClass(Info);
```
kOchirasu, line 25: `writer.WriteClass<PlayerInfo>(Info);`

Maple2.Server.Core/proto/channel/channel.proto
```diff
@@ -105,3 +107,52 @@ message PartyRequest {
 message PartyResponse {
   int32 error = 1;
 }
+
+message GroupChatRequest {
```
kOchirasu, line 111: This seems duplicated, you can put in `common.proto` and then import it.

See: `ChatRequest`/`BuddyRequest`

Maple2.Server.Game/Manager/GroupChatManager.cs
```diff
@@ -0,0 +1,181 @@
...
+    private void BeginListen(GroupChatMember member) {
+        // Clean up previous token if necessary
+        if (member.TokenSource != null) {
+            logger.Warning("BeginListen called on Member {Id} that was already listening", member.CharacterId);
+            EndListen(member);
+        }
+
+        member.TokenSource = CancellationTokenSource.CreateLinkedTokenSource(tokenSource.Token);
+        CancellationToken token = member.TokenSource.Token;
+        var listener = new PlayerInfoListener(UpdateField.Party, (type, info) => SyncUpdate(token, member.CharacterId, type, info));
```
kOchirasu, line 153: `UpdateField.Party` is probably wrong here, you only want to subscribe to the update events that matter to group chat.

Maple2.Server.Game/Manager/GroupChatManager.cs
```diff
@@ -0,0 +1,181 @@
...
+
+    private bool SyncUpdate(CancellationToken cancel, long characterId, UpdateField type, IPlayerInfo info) {
+        if (cancel.IsCancellationRequested || groupChat == null || !groupChat.Members.TryGetValue(characterId, out GroupChatMember? member)) {
+            return true;
+        }
+
+        bool wasOnline = member.Info.Online;
+        member.Info.Update(type, info);
+
+        //session.Send(PartyPacket.Update(member));
```
kOchirasu, line 171: remove? Do you need something else here?

Maple2.Server.Game/PacketHandlers/GroupChatHandler.cs
```diff
@@ -0,0 +1,151 @@
...
+        var error = (GroupChatError) response.Error;
+        if (error != GroupChatError.none) {
+            session.Send(GroupChatPacket.Error(error, session.PlayerName, string.Empty));
+            return;
+        }
+
+        if (!session.GroupChats.TryAdd(response.GroupChat.Id, new GroupChatManager(response.GroupChat, session))) {
+            throw new InvalidOperationException($"Failed to add group chat: {response.GroupChat.Id}");
+        }
+        ;
```
kOchirasu, line 71: nit: remove `;`

Maple2.Server.Game/Packets/GroupChatPacket.cs
```diff
@@ -0,0 +1,134 @@
...
+        Error = 13,
+    }
+
+    public static ByteWriter Load(GroupChat groupChat) {
+        var pWriter = Packet.Of(SendOp.GroupChat);
+        pWriter.Write<Command>(Command.Load);
+        pWriter.WriteInt(groupChat.Id);
+        pWriter.WriteByte((byte) groupChat.Members.Count);
+        foreach ((long characterId, GroupChatMember member) in groupChat.Members) {
+            pWriter.WriteByte(1); // Login notification ?
```
kOchirasu, line 32: should this be a bool then? Also if that's the case does it mean you get spammed with "login" for each member of the chat whenever it loads?

---

Seems like this is probably the same byte as `AddMember` maybe it just means there IS a member.

Maple2.Server.Game/Packets/GroupChatPacket.cs
```diff
@@ -0,0 +1,134 @@
...
+            pWriter.WriteClass(member.Info);
```
kOchirasu, line 33: For clarity, always include type with `WriteClass<>`

Maple2.Server.Game/Packets/GroupChatPacket.cs
```diff
@@ -0,0 +1,134 @@
...
+        return pWriter;
+    }
+
+    public static ByteWriter AddMember(PlayerInfo targetCharacter, string inviterName, int groupChatId) {
+        var pWriter = Packet.Of(SendOp.GroupChat);
+        pWriter.Write<Command>(Command.AddMember);
+        pWriter.WriteInt(groupChatId);
+        pWriter.WriteUnicodeString(inviterName);
+        pWriter.WriteByte(1); // Unknown
+        pWriter.WriteClass(targetCharacter);
```
kOchirasu, line 81: same as above

Maple2.Server.Core/proto/channel/channel.proto
Zintixx: Only some overlap. (ex. Create in world only or AddMember in channel only). Should I move the duplicated ones to common?

kOchirasu: Doesn't moving it to common make it easier to forward the request (since you don't need to translate between two protos) if so I think it makes sense even if it's only a subset.

PR #66, opened 2023-10-31
Performance Ensemble

Zintixx:
- Also Fixes the write party packet
- Also auto disbanding parties if nobody is online.

### DIFF
--- a/Maple2.Model/Game/Party/Party.cs
+++ b/Maple2.Model/Game/Party/Party.cs
@@ -59,26 +59,5 @@ public class Party : IByteSerializable {
         writer.WriteBool(false); // unk bool
         writer.WriteInt(DungeonId);
         writer.WriteBool(false); // unk bool
-        writer.WriteByte();
-
-        //TODO: Listed Party
-        bool listedParty = false;
-        writer.WriteBool(listedParty);
-        if (listedParty) {
-            /*
-            writer.WriteLong(PartyFinderId);
-            writer.WriteInt(Id);
-            writer.WriteInt();
-            writer.WriteInt();
-            writer.WriteUnicodeString(Name);
-            writer.WriteBool(Approval);
-            writer.WriteInt(party.Members.Count);
-            writer.WriteInt(party.RecruitMemberCount);
-            writer.WriteLong(party.Leader.AccountId);
-            writer.WriteLong(party.Leader.CharacterId);
-            writer.WriteUnicodeString(party.Leader.Name);
-            writer.WriteLong(party.CreationTimestamp);
-            */
-        }
     }
 }
--- a/Maple2.Model/Game/Party/Party.cs
+++ b/Maple2.Model/Game/Party/Party.cs
@@ -49,8 +49,7 @@ public class Party : IByteSerializable {
         writer.WriteInt(Id);
         writer.WriteLong(LeaderCharacterId);
 
-        byte memberCount = (byte) Members.Count;
-        writer.WriteByte(memberCount);
+        writer.WriteByte((byte) Members.Count);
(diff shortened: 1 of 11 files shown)


Maple2.Model/Game/Party/Party.cs
```diff
@@ -60,5 +59,26 @@ public class Party : IByteSerializable {
         writer.WriteBool(false); // unk bool
         writer.WriteInt(DungeonId);
         writer.WriteBool(false); // unk bool
+        writer.WriteByte();
+
+        //TODO: Listed Party
```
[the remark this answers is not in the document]
Zintixx: oop. fixed

Maple2.Server.Game/Manager/PartyManager.cs
```diff
@@ -173,6 +173,22 @@ public class PartyManager : IDisposable {
...
     }
 
+    public bool CheckDisband(long characterId) {
+        if (Party == null) {
+            return false;
+        }
+
+        // Check if any other player is online
+        if (!Party.Members.Values.Any(partyMember => partyMember.Info.Online && partyMember.CharacterId != characterId)) {
+            world.Party(new PartyRequest {
```
kOchirasu, line 183: why does this only need to be sent if someone is online? I feel like in all cases, the world server needs to know that the party should no longer exist since it's the source of truth.

Maple2.Server.Game/PacketHandlers/InstrumentHandler.cs
```diff
@@ -172,9 +172,58 @@ private enum Command : byte {
         session.Instrument = null;
     }
 
-    private void HandleJoinEnsemble(GameSession session, IByteReader packet) { }
+    private void HandleJoinEnsemble(GameSession session, IByteReader packet) {
```
kOchirasu, line 175: I have no idea how ensembles work anymore, but it does seem a bit strange to be that `JoinEnsemble` affects everyone in your party, but `LeaveEnsemble` only affects yourself. Is that correct?

Maple2.Server.Game/Manager/PartyManager.cs
Zintixx: It does the opposite. It's sent only if nobody is online.

Maple2.Server.Game/PacketHandlers/InstrumentHandler.cs
Zintixx: Yeah, it's correct. I think it might change a character's sync flags. If the packet was broadcasted to the field, it stops everyone from playing.

Maple2.Server.Game/Manager/PartyManager.cs
kOchirasu: okay but if nobody is online, the world server still needs to know right? otherwise the people would still be in the party when they logged back on.

Zintixx: I'm not sure I understand.
The function checks if nobody is online. If nobody is online, it sends a World request to disband and remove the party from the game.
The party no longer exists and there's no way someone could login with the party anymore.

Maple2.Server.Game/Session/GameSession.cs
```diff
@@ -441,7 +441,9 @@ public sealed partial class GameSession : Core.Network.Session {
 #endif
             Guild.Dispose();
             Buddy.Dispose();
-            Party.Dispose();
+            if (!Party.CheckDisband(CharacterId)) {
```
kOchirasu, line 444: I think `Dispose()` should always happen as it's cleaning up resources, maybe `CheckDisband` in `Dispose()`?

